PR #8, opened 2020-09-07
add sermon list page

narthur:


lib/api.ts
```diff
@@ -98,3 +100,23 @@ fragment SermonFragment on Recording {
 	);
 	return data && data.sermon;
 }
+
+export async function getSermonCount(language): Promise<number> {
+	const data = fetchAPI(
+		`
+	query countQuery($language: Language!) {
```
mattleff, line 107: It would be nice if we could get the GraphQL queries auto-formatted.

lib/constants.ts
```diff
@@ -1,3 +1,5 @@
+export const entriesPerPage = 25;
```
mattleff, line 1: I wonder if we should use upper case snake case for the consts:

```suggestion
export const ENTRIES_PER_PAGE = 25;
```

pages/[language]/sermons/[id].ts
```diff
@@ -1,5 +1,5 @@
 import SermonDetail from '../../../containers/sermon/detail';
-import { getRecentSermons as getLatestSermons, getSermon } from '../../../lib/api';
+import { getSermon, getSermons } from '../../../lib/api';
```
Comment (mattleff) [inline pages/[language]/sermons/[id].ts:2]: We should add some tsconfig paths to avoid these deep parent paths.

lib/api.ts
```diff
@@ -24,11 +26,11 @@ async function fetchAPI(query, { variables } = {}) {
 	return json.data;
 }
 
-export async function getRecentSermons(language) {
+export async function getSermons(language, { offset = null, first = 1000 } = {}) {
 	const data = await fetchAPI(
 		`
-  query loadPagesQuery($language: Language!, $cursor: String, $first: Int!) {
-    sermons(language: $language, first: $first, after: $cursor, orderBy: {direction: DESC, field: CREATED_AT}) {
+  query loadPagesQuery($language: Language!, $offset: Int, $first: Int!) {
```
mattleff, line 32: Should the query name match the function name?

containers/sermon/detail.tsx
```diff
@@ -1,51 +1,50 @@
...
+	if (!router.isFallback && !sermon) {
+		return <ErrorPage statusCode={404} />;
+	}
+
+	const imageSrc = sermon.imageWithFallback.url,
+		imageAlt = sermon.title;
+
+	return (
+		<div>
+			{router.isFallback ? (
```
mattleff, line 19: It might be good to extract a loading container (maybe even HOC?) that would be shown client-side while waiting for data in the fallback state.

mattleff (approved):
Great work! Just a few comments, none of which are blocking. :shipit:

lib/api.ts
narthur: Punting on this one.

PR #12, opened 2020-09-14
work on ts types; add sermon list pagination

narthur:


types.d.ts
```diff
@@ -1,11 +1,16 @@
 interface Sermon {
```
mattleff, line 1: Eventually we should generate these from the GraphQL queries

mattleff (approved):
Looks great! 👍

PR #15, opened 2020-09-17
Testmonies

narthur:


lib/helpers.ts
```diff
@@ -0,0 +1,21 @@
+import { ENTRIES_PER_PAGE, LANGUAGES } from '@lib/constants';
+
+export const getNumberedStaticPaths = async (
+	sectionSegments: string,
+	getCount: (language: string) => Promise<number>
+) => {
+	const pathSetPromises = Object.keys(LANGUAGES).map(async (k) => {
+			const sermonCount = await getCount(k),
```
mattleff, line 10: `entryCount` or `itemCount`?

lib/api.ts
```diff
@@ -123,3 +123,57 @@ export async function getSermonCount(language: string): Promise<number> {
 	);
 	return _.get(data, 'sermons.aggregate.count') || 0;
 }
+
+export async function getTestimonyCount(language: string): Promise<number> {
```
mattleff, line 127: I feel like having all our queries in `lib/api.ts` is going to get cumbersome at some point. Maybe we use `lib/api/*` and re-export everything in `lib/api.ts`?

lib/api.ts
```diff
@@ -123,3 +123,57 @@ export async function getSermonCount(language: string): Promise<number> {
...
+  `,
+		{
+			variables: {
+				language,
+				offset,
+				first,
+			},
+		}
+	);
+	return data && data.testimonies;
```
mattleff, line 178:
```suggestion
	return data?.testimonies;
```

lib/helpers.ts
```diff
@@ -0,0 +1,48 @@
...
+				numbers = Array.from(Array(pageCount).keys()),
+				base = LANGUAGES[k].base_url;
+
+			return numbers.map((x) => `/${base}/${sectionSegments}/${x + 1}`);
+		}),
+		pathSets = await Promise.all(pathSetPromises);
+
+	return {
+		paths: pathSets.flat(),
+		fallback: 'unstable_blocking',
```
mattleff, line 21: `fallback: 'unstable_blocking',` doesn't seem to have anything to do with `getNumberedStaticPaths()`. I wonder if this function should return an array of paths and we use another function to generate:

```
{
	paths,
	fallback: 'unstable_blocking',
}
```

lib/helpers.ts
```diff
@@ -0,0 +1,48 @@
...
+	};
+};
+
+export async function getPaginatedStaticProps(language: string, pageIndex: number, getter: any) {
```
mattleff: There might be a more specific type but you could start with:

```suggestion
export async function getPaginatedStaticProps(language: string, pageIndex: number, getter: Promise<{ nodes: any[], aggregate: { count: number; } }>) {
```

pages/[language]/sermons/page/[i].tsx
```diff
@@ -16,42 +14,11 @@ interface GetStaticPropsArgs {
...
-		props: {
-			sermons,
-			pagination: {
-				total,
-				current: parseInt(i),
-			},
-		},
-		revalidate: 10,
-	};
+	return await getPaginatedStaticProps(language, parseInt(i), getSermons);
```
Comment (mattleff) [inline pages/[language]/sermons/page/[i].tsx:19]: You don't need to `await` before returning (it's considered a [lint issue](https://eslint.org/docs/rules/no-return-await)).

mattleff (approved):
Great work! 👍

PR #17, opened 2020-09-24
Style testimonies

narthur:


containers/testimonies.module.scss
```diff
@@ -0,0 +1,47 @@
...
+  margin: 0;
+}
+
+.quotation {
+  color: $gray-mid;
+  padding: 1em;
+  background-color: $gray-bleached;
+  border-radius: 10px;
+  position: relative;
+  margin: 0 0 10px 0;
```
mattleff, line 21: Is there a reason why margin here is in `px` and on `.source` you're using `em`?

mattleff (approved):
Looks good to me!

containers/testimonies.module.scss
narthur: Good point. Turns out this margin wasn't needed at all.

PR #18, opened 2020-09-29
Format intl

narthur:


components/organisms/testimonies.tsx
```diff
@@ -0,0 +1,38 @@
...
+		: query.language || 'en';
+
+	return (
+		<IntlProvider
+			messages={messages[language]}
+			locale={language}
+			defaultLocale="en"
+		>
+			<h2>
+				<FormattedMessage id={'title'} defaultMessage={'Testimonies'} />
```
mattleff, line 32: I'd also use [babel-plugin-react-intl](https://formatjs.io/docs/guides/bundler-plugins) to avoid having to define ids explicitly.

components/organisms/testimonies.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { IntlProvider, FormattedMessage, FormattedNumber } from 'react-intl';
+import { useRouter } from 'next/router';
+
+type Messages = {
+	[langCode: string]: {
+		[key: string]: string;
+	};
+};
+
+const messages: Messages = {
```
mattleff, line 11: You can extract the messages automatically with https://formatjs.io/docs/getting-started/message-extraction.

components/organisms/testimonies.tsx
mattleff: Also, if you set up [eslint-plugin-formatjs](https://formatjs.io/docs/tooling/linter) you can enforce having a default message with `enforce-default-message`. Probably most or all of the rules they have are good.

src/styles/Home.module.css
```diff
@@ -0,0 +1,123 @@
...
+	justify-content: center;
+	align-items: center;
+}
+
+.title a {
+	color: #0070f3;
+	text-decoration: none;
+}
+
+.title a:hover,
```
mattleff, line 43: If this is changed to be a `.scss` I'd recommend nesting these styles in the `.title a` above.

mattleff (approved):
Looks good! 👍

src/styles/Home.module.css
```diff
@@ -0,0 +1,123 @@
+.container {
```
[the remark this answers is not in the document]
narthur: This is a default Next file which will eventually be discarded.

PR #23, opened 2020-10-20
sermon list images

narthur:


src/components/molecules/recordingList.spec.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+describe('recording list', () => {
+	it('renders', async () => {
+		await renderComponent();
+	});
+
+	it('has image', async () => {
+		const { getByRole } = await renderComponent();
+
+		// broken type is annoying. Shouldn't require container since it's being
```
mattleff, line 32: What does this mean? Is there a GH issue we could link to here?

mattleff (approved):
Looks good!

PR #24, opened 2020-10-21
speaker names in sermons list

narthur:


next.config.js
```diff
@@ -4,4 +4,13 @@ module.exports = withPWA({
 	pwa: {
 		dest: 'public',
 	},
+	async redirects() {
```
mattleff, line 7: Does this need to be `async`?

src/containers/testimonies.module.scss
```diff
@@ -1,6 +1,6 @@
 @import '../styles/common';
 
-.list {
+.table {
```
mattleff, line 3: This shouldn't be renamed...

src/components/molecules/recordingListEntry.tsx
```diff
@@ -0,0 +1,42 @@
...
+				</a>
+			</td>
+			<td>
+				<a href={`/${lang}/sermons/${sermon.id}`}>{sermon.title}</a>
+			</td>
+			<td className={styles.presenters}>
+				<ul>
+					{persons.map((p: Person, i: number) => {
+						return (
+							<li key={i}>
```
mattleff, line 33: You could use `p.id` here since that should be unique.

mattleff (approved):
LGTM! 🧑‍🦱

next.config.js
narthur: No, but I think I'll leave this for now so I know how it should work once we start doing dynamic redirects.

PR #25, opened 2020-10-21
add duration to recordings list

narthur:


narthur:
https://audioverse-next-git-sermon-list-speakers.avorg.vercel.app/en/sermons/page/1

src/components/molecules/recordingListEntry.tsx
```diff
@@ -4,6 +4,22 @@ import React from 'react';
 
 import styles from '@components/molecules/recordingListEntry.module.scss';
 
+const formatDuration = (seconds: number, m = 0, h = 0): string => {
+	seconds = Math.round(seconds);
+
+	if (seconds >= 60 * 60) {
+		return formatDuration(seconds - 60 * 60, m, h + 1);
```
mattleff, line 11: I'm not a fan of the recursive solution here. Calling `formatDuration(3554)` recursively calls `formatDuration()` 60 times. How about something like this (haven't tested this):

```ts
const formatDuration = (duration: number): string => {
     duration = Math.round(duration);
     const hours = Math.floor(duration / 3600);
     const minutes = Math.floor((duration - (hours * 3600)) / 60);
     const seconds = String(duration % 60).padStart(2, '0');
     return `${hours ? hours + ':' + String(minutes).padStart(2, '0') : minutes}:${seconds}`;
 };
```

mattleff (approved):
Looks good other than one comment 🏎️

src/components/molecules/recordingListEntry.tsx
narthur: Nice, works great!

PR #26, opened 2020-10-22
add filter ui to recording list [old]

narthur:
just making this pr to test

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,8 @@
   },
   "dependencies": {
     "@types/lodash": "^4.14.161",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -16,14 +16,17 @@
   },
   "dependencies": {
     "@types/lodash": "^4.14.161",
+    "@xstate/react": "^1.0.1",
+    "@xstate/test": "^0.4.1",
```
mattleff, line 20: I think this should be a dev dependency.

mattleff (approved):
Looks good! 👍

PR #27, opened 2020-10-28
List video filter ssr

narthur:


src/components/molecules/recordingList.tsx
```diff
@@ -11,11 +12,22 @@ interface RecordingListProps {
...
+
 	return (
-		<table className={styles.list}>
-			{sermons.map((s) => (
-				<RecordingListEntry key={s.id} sermon={s} />
-			))}
-		</table>
+		<div>
+			<div>
+				<a href={`/${lang}/sermons/all/page/1`}>All</a>
```
mattleff, line 20: I wonder if we could take a `basePath` prop to avoid having to define the path here. Wdyt?

mattleff (approved):
Looks good! 👍

src/components/molecules/recordingList.tsx
narthur: Yup! I moved this to the page level so it doesn't have to be abstracted.

PR #28, opened 2020-11-03
All sermons RSS feed

narthur:


src/pages/[language]/sermons/all/page/[i].tsx
```diff
@@ -17,9 +19,25 @@ interface GetStaticPropsArgs {
...
+		base_url,
+		parseInt(i),
+		getSermons
+	);
+
+	if (i === '1' && response.props.nodes) {
+		await createFeed({
+			recordings: response.props.nodes,
+			projectRelativePath: `public/${base_url}/sermons/all.xml`,
+			title: `AudioVerse Recent Recordings: ${lang.display_name}`,
```
Comment (mattleff) [inline src/pages/[language]/sermons/all/page/[i].tsx:36]: Does this title need to be internationalized?

src/lib/createFeed.ts
```diff
@@ -0,0 +1,51 @@
...
+	title,
+	projectRelativePath,
+}: CreateFeedProps): Promise<void> {
+	const feed = new Feed({
+		id: '',
+		title,
+		copyright: '',
+	});
+
+	recordings.map((i) => {
```
mattleff, line 26: This is a nit, but should `i` be either `r` or `recording`? `i` feels like an index.

package.json
```diff
@@ -16,8 +16,11 @@
   },
   "dependencies": {
     "@types/lodash": "^4.14.161",
+    "@types/mkdirp": "^1.0.1",
+    "feed": "^4.2.1",
     "full-icu": "^1.3.1",
     "lodash": "^4.17.20",
+    "mkdirp": "^1.0.4",
```
mattleff, line 23: Will you be using `mkdirp`?

src/lib/createFeed.ts
```diff
@@ -0,0 +1,51 @@
...
+		const file = _.get(i, 'audioFiles[0]') || _.get(i, 'videoFiles[0]'),
+			url = _.get(file, 'url'),
+			length = _.get(file, 'filesize');
+
+		if (!url) return;
+
+		feed.addItem({
+			title: i.title,
```
mattleff, line 34: It would be good to add `TODO`s here for the elements that aren't here yet.

mattleff (approved):
Looks good 👍

src/lib/getIntlMessages.ts
```diff
@@ -0,0 +1,9 @@
+import { PROJECT_ROOT } from '@lib/constants';
+
+export default function getIntlMessages(language: string): any {
+	try {
+		return require(`${PROJECT_ROOT}/public/compiled-lang/${language}.json`);
```
mattleff, line 5: Would it be better to have a whitelist of supported languages instead of try-catching?

src/pages/[language]/sermons/all/page/[i].tsx
```diff
@@ -29,11 +32,29 @@ export async function getStaticProps({
 		getSermons
 	);
 
+	const messages = getIntlMessages(base_url);
```
Comment (mattleff) [inline src/pages/[language]/sermons/all/page/[i].tsx:35]: I wonder if we could make a method that wraps all this boilerplate so all that needs to be provided is lines 43-47/50. Wdyt?

mattleff (approved):
LGTM! 👍

src/lib/createFeed.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export default async function createFeed({
+	recordings,
+	title,
+	projectRelativePath,
+}: CreateFeedProps): Promise<void> {
+	const feed = new Feed({
+		id: '',
+		title,
+		copyright: '',
```
mattleff, line 23: Are there any elements still TODO at the feed level?

src/pages/[language]/sermons/all/page/[i].tsx
Comment (narthur) [reply src/pages/[language]/sermons/all/page/[i].tsx:35]: I don't think I can extract most of this because the `intl.formatMessage` call needs to be retained for `formatjs extract` to work.

Comment (mattleff) [reply src/pages/[language]/sermons/all/page/[i].tsx:35]: Hmm, good point. How about something like this:

```ts
const intl = getIntl(base_url);
const title = intl.formatMessage( ... );
```

PR #29, opened 2020-11-06
add rss links to sermon list page

narthur:


src/containers/sermon/list.tsx
```diff
@@ -15,8 +17,15 @@ export interface SermonListProps {
...
+	const filter = _.get(match, '[1]', 'all');
+
 	return (
 		<div>
+			<a href={`/${lang}/sermons/${filter}.xml`} target={'_blank'}>
+				RSS
+			</a>
 			<div>
 				<a href={`/${lang}/sermons/all/page/1`}>All</a>
 				<a href={`/${lang}/sermons/video/page/1`}>Video</a>
```
mattleff, line 31: This is not related to this PR but line 38 should use `lang` instead of `en`.

mattleff (approved):
👍 I was wondering if we also want to add the `rel=alternate` meta tag for the RSS?

src/containers/sermon/list.tsx
```diff
@@ -23,7 +23,11 @@ function SermonList({ nodes, pagination }: SermonListProps) {
 
 	return (
 		<div>
-			<a href={`/${lang}/sermons/${filter}.xml`} target={'_blank'}>
+			<a
+				href={`/${lang}/sermons/${filter}.xml`}
+				target={'_blank'}
+				rel={'noreferrer noopener'}
```
mattleff, line 29: Why do we need these? I'm not saying they're wrong, just wondering why they're needed.

narthur: It's prob not necessary here since it's an internal link, but tests are configured to fail if a `target="_blank"` link doesn't have `noreferrer`, and I don't think it creates any issues to have it when it isn't needed.

mattleff: Yeah, that's fine. It will affect our access logs, so we wouldn't be able to break out how many people are clicking through to the RSS from their browser. That kind of stat could be helpful to know whether we should invest in a XML stylesheet. OTOH, if we know we're going to do that anyway, I can't think of any other actionable difference missing the referrer would make.

PR #30, opened 2020-11-20
style recording list

narthur:


mattleff (approved):
LGTM!

src/components/molecules/favorite.spec.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+		jest.spyOn(api, 'setFavorited').mockRejectedValue('error');
+
+		userEvent.click(button);
+
+		await expect(findByText('Unfavorite')).resolves.toBeInTheDocument();
+		await expect(findByText('Favorite')).resolves.toBeInTheDocument();
+	});
+
+	it('tests right', async () => {
```
mattleff, line 93: Need this?

src/components/molecules/favorite.spec.tsx
```diff
@@ -0,0 +1,128 @@
...
+		const mock = jest.fn();
+		mock.mockReturnValue(true);
+		mock.mockReturnValue(false);
+		expect(mock()).toBeFalsy();
+	});
+
+	it('tests isFavorited right', async () => {
+		const isFavoritedSpy = jest.spyOn(api, 'isFavorited');
+		isFavoritedSpy.mockResolvedValue(false);
```
mattleff, line 102: Do you need both `.mockResolvedValue()` calls?

src/components/molecules/favorite.spec.tsx
```diff
@@ -0,0 +1,128 @@
...
+		isFavoritedSpy.mockResolvedValue(true);
+		expect(isFavorited(-1)).toBeTruthy();
+	});
+
+	it('tests throwing properly', async () => {
+		const { findByText } = await renderComponent();
+
+		await expect(findByText('Unfavorite')).rejects.toThrow();
```
mattleff, line 110: What is this testing?

src/components/molecules/favorite.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+import { useIsFavorited } from '@lib/api';
+
+export default function Favorite({ id }: { id: string }): JSX.Element {
+	const { isFavorited, toggle } = useIsFavorited(id);
+
+	return (
+		<button
+			onClick={async () => {
```
mattleff, line 10: Could/should this be `onClick={toggle}`?

src/lib/api/setFavorited.ts
```diff
@@ -0,0 +1,24 @@
...
+
+// TODO: Replace `any` in return type with something more specific
+export async function setFavorited(
+	id: number | string,
+	favorite: boolean
+): Promise<any> {
+	const variables = { id },
+		query = favorite ? favoriteQuery : unfavoriteQuery;
+
+	return await fetchApi(query, { variables });
```
mattleff, line 23: Is there a reason to `return await fetchApi(...)` instead of `return fetchApi(...)`?

src/lib/api/useIsFavorited.ts
```diff
@@ -0,0 +1,38 @@
...
+	const { data: faved } = useQuery(queryKey, async () => isFavorited(id));
+
+	const [toggle] = useMutation(
+		() => {
+			// TODO: Consider using setFavorited response to update data instead of
+			//  invalidating `isFavorited` query
+
+			// onMutate optimistically changes cached data, so we don't
+			// flip faved since it's already been flipped in onMutate
+			return setFavorited(id, !!faved);
```
mattleff, line 21: Should this be `!faved`?

src/lib/api/useIsFavorited.ts
```diff
@@ -0,0 +1,38 @@
...
+import { isFavorited } from '@lib/api/isFavorited';
+import { setFavorited } from '@lib/api/setFavorited';
+
+export function useIsFavorited(
+	id: string
+): { isFavorited: boolean | undefined; toggle: () => void } {
+	const cache = useQueryCache();
+	const queryKey = ['isFavorited', id];
+
+	const { data: faved } = useQuery(queryKey, async () => isFavorited(id));
```
mattleff, line 12: I don't think you need to mark the closure `async` since you aren't `await`-ing anything.

mattleff (approved):
Looks good overall 👍

src/components/molecules/favorite.spec.tsx
narthur: You found all my "I need to verify everything because I don't know what anything's doing anymore" tests 😆

src/components/molecules/favorite.tsx
narthur: types complain when I try

src/lib/api/useIsFavorited.ts
narthur: No, because the optimistic update will have already flipped the value by then

src/components/molecules/favorite.tsx
mattleff: It doesn't complain when I try it locally. What error is it giving you?

src/lib/api/useIsFavorited.ts
narthur: It makes it easier to test since it will always return a promise even if a return value hasn't been defined on a mock...

src/lib/api/useIsFavorited.ts
```diff
@@ -0,0 +1,38 @@
...
+			// TODO: Consider using setFavorited response to update data instead of
+			//  invalidating `isFavorited` query
+
+			// onMutate optimistically changes cached data, so we don't
+			// flip faved since it's already been flipped in onMutate
+			return setFavorited(id, !!faved);
+		},
+		{
+			onMutate: () => {
+				cache.cancelQueries(queryKey);
```
mattleff, line 25: Do we want to use the [`exact`](https://react-query.tanstack.com/docs/api/#querycachecancelqueries) option?

src/lib/api/useIsFavorited.ts
mattleff: Ok, I should've read the comment more closely. 🙃 It still feels like a unusual flow... Maybe it would be better with [mutation variables](https://react-query.tanstack.com/docs/guides/mutations#mutation-variables) pinning to the previous `faved` value? Btw, wdyt about renaming `faved` to `isFavorited`?

narthur: You're right, it is awkward... I'll see what I can do.

The reason it isn't `isFavorited` is because that's already the name of a function I'm importing.

mattleff: I wonder if the import should be `isRecordingFavorited` since theoretically we'll have `isPersonFavorited` at some point.

src/components/molecules/favorite.tsx
narthur: I guess something changed... now it's not complaining

src/lib/api/useIsFavorited.ts
narthur: I'm going to stick with the awkward flow for now. I don't think it's worth the time to try to fix it right now.

PR #31, opened 2020-12-10
Login

narthur:


src/components/molecules/recordingList.tsx
```diff
@@ -2,6 +2,8 @@ import React from 'react';
 
 import RecordingListEntry from '@components/molecules/recordingListEntry';
 
+import type { Sermon } from '../../../types';
```
mattleff, line 5: Might be nice to setup a path in `tsconfig.json` for like `@types` to simplify these imports.

src/lib/api/fetchApi.ts
```diff
@@ -1,10 +1,27 @@
...
 export async function fetchApi(
 	query: string,
 	{ variables = {} } = {}
 ): Promise<any> {
-	const headers = { 'Content-Type': 'application/json' };
+	const cookies = getCookies(_request);
+
+	const headers = {
+		'Content-Type': 'application/json',
+		'x-av-session': _.get(cookies, 'avSession'),
```
mattleff, line 23: Does `lodash.get` buy us anything over `cookies?.avSession`?

src/lib/api/login.ts
```diff
@@ -0,0 +1,25 @@
...
+`;
+
+import { fetchApi } from '@lib/api/fetchApi';
+
+export async function login(email: string, password: string): Promise<boolean> {
+	const variables = { email, password },
+		data = await fetchApi(query, { variables }),
+		token = data.login.authenticatedUser.sessionToken;
+
+	if (token) {
```
mattleff, line 20: `token` should always be set, right? Because if the login failed `fetchApi` will throw and the exception will be passed back to the callback of `login`, right?

src/lib/getCookies.ts
```diff
@@ -0,0 +1,19 @@
...
+
+// TODO: Improve req type
+export default function getCookies(
+	req: IncomingMessage | null
+): { [key: string]: string } {
+	if (req) {
+		return cookie.parse(_.get(req.headers, 'cookie') || '');
+	}
+
+	if (process.browser) {
```
mattleff, line 14: What if `process` doesn't exist?

src/pages/_app.tsx
```diff
@@ -13,14 +16,20 @@ function MyApp<P>({
 	Component: typeof React.Component;
 	pageProps: P;
 }): JSX.Element {
+	const queryCache = new QueryCache();
+
 	return (
 		<div className={'template-base'}>
 			<Head>
 				<title>AudioVerse</title>
```
mattleff, line 24: This is off-topic, but how are we going to handle setting head data on a per-page basis?

src/pages/[language]/profile.tsx
```diff
@@ -0,0 +1,34 @@
...
+	props: {
+		dehydratedState: DehydratedState;
+	};
+}
+
+export async function getServerSideProps({
+	req,
+}: {
+	req: IncomingMessage;
+}): Promise<ServerSideProps> {
```
Comment (mattleff) [inline src/pages/[language]/profile.tsx:22]: I think we could improve the types here:

image

mattleff (approved):
Great work! 👍

src/pages/_app.tsx
narthur: I think you just add `<Head />` elements to multiple components and it's all gathered for that tree.

PR #32, opened 2020-12-11
video.js

narthur:


src/lib/api/getSermons.ts
```diff
@@ -28,11 +28,13 @@ fragment SermonsFragment on Recording {
 		url
 		duration
 		filesize
+		mimeType
 	}
 	videoFiles {
```
mattleff, line 33: Do you need all these fields for the sermons list query?

src/components/molecules/recordingListEntry.tsx
```diff
@@ -19,8 +19,9 @@ export default function RecordingListEntry({
 }: {
 	sermon: Sermon;
 }): JSX.Element {
-	const lang = useLanguage(),
-		persons: Person[] = _.get(sermon, 'persons', []);
+	const lang = useLanguage();
+	const persons: Person[] = _.get(sermon, 'persons', []);
+	const videoFiles: VideoFile[] = _.get(sermon, 'videoFiles', []);
```
mattleff: Is `VideoFile` imported?

src/components/molecules/recordingListEntry.tsx
```diff
@@ -49,6 +50,7 @@ export default function RecordingListEntry({
 				</ul>
 			</td>
 			<td className={styles.duration}>{formatDuration(sermon.duration)}</td>
+			<td>{videoFiles.length > 0 ? 'Video' : 'Audio'}</td>
```
mattleff, line 53: Could I add `hasVideo` & `hasAudio` fields that would make this cleaner?

src/lib/api/getSermons.ts
avorg: For the purpose of keeping it more-or-less consistent with my manual types I think it's fine for now.

mattleff (approved):
Looks good! ☃️

src/components/molecules/recordingListEntry.tsx
narthur: That's  a type that I renamed but forgot to change here. It's changed now.

src/components/molecules/recordingListEntry.tsx
narthur: Maybe that would be good. My only question is if we know how those fields should be defined? Should they include streams or just videos for hasVideo, for instance?

mattleff: It would be if there's a video source (which would indicate the transcoded variations) or not.

src/lib/api/getSermons.ts
mattleff: Ok. It will slow down API response time + increase response size, but I guess we can "make it fast"-er later if we need to.

PR #35, opened 2021-01-05
add sermon detail tags

narthur:


types.d.ts
```diff
@@ -56,3 +63,10 @@ interface StaticPaths {
 	paths: string[];
 	fallback: boolean;
 }
+
+interface RecordingTag {
```
mattleff, line 67: It would be nice to generate these from the queries.

mattleff (approved):
LGTM! :shipit:

types.d.ts
narthur: @mattleff Do you think now is a good time to do that? How would I go about it?

mattleff: I think the sooner the better. You could use GraphQL Code Generator with a couple of their plugins: https://graphql-code-generator.com/docs/plugins/typescript-react-query/

PR #36, opened 2021-01-06
add graphql codegen

narthur:


narthur:
@mattleff FYI this PR is now ready for review.

getter-plugin.js
```diff
@@ -0,0 +1,32 @@
+module.exports = {
+	plugin: (schema, documents, config, info) => {
+		const result = documents
+			.map((doc) => {
+				return doc.document.definitions
+					.filter((def) => {
+						return (
+							def.kind !== 'FragmentDefinition' && def?.operation !== 'mutation'
```
mattleff, line 8: We don't need `def?.operation` since `def` should always be defined.

src/jest.spec.ts
```diff
@@ -0,0 +1,36 @@
+import { when } from 'jest-when';
```
mattleff, line 1: Do we need this file?

mattleff (approved):
Great work! It's really nice to have those types generated :shipit: 🍹

src/jest.spec.ts
narthur: Yes, I'd like to keep this around. If an API is confusing enough that I feel the need to write a test to figure it out, I like to keep the test for documentation purposes.

PR #37, opened 2021-01-18
add series link to sermon detail page

narthur:


src/containers/sermon/detail.tsx
```diff
@@ -182,6 +184,20 @@ function SermonDetail({ sermon }: SermonDetailProps) {
...
+			{sermon?.sequence && (
+				<>
+					<h2>
+						<FormattedMessage
+							id="sermonDetailPage__seriesTitle"
+							defaultMessage="Series"
+							description="Sermon detail series title"
+						/>
+					</h2>
+					<a href={`/${langRoute}/series/${sermon?.sequence?.id}`}>
```
mattleff, line 196: I think it would be good to have a URL formatting class/functions so if we need to change the URL for an entity type we have a single place to do that.

mattleff (approved):
LGTM 👍🏻

package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "dev": "next dev",
-    "build": "next build",
+    "build": "NODE_ICU_DATA=node_modules/full-icu next build",
```
mattleff, line 7: Do we need this for `start` too, or does SSR not trigger that code path?

mattleff (approved):
Looks good!

PR #38, opened 2021-01-20
Sermon detail page

narthur:


src/components/molecules/recordingList.tsx
```diff
@@ -6,17 +6,19 @@ import { RecordingListFragment } from '@lib/generated/graphql';
...
+	recordings,
 }: RecordingListProps): JSX.Element {
 	return (
 		<table className={styles.list}>
 			<tbody>
-				{sermons &&
-					sermons.map((s, i) => <RecordingListEntry key={i} sermon={s} />)}
+				{recordings &&
+					recordings.map((rec, i) => (
+						<RecordingListEntry key={i} recordings={rec} />
```
mattleff, line 20: Using `rec.id` for the `key` would be more stable than `i`.

src/components/organisms/recording.tsx
```diff
@@ -0,0 +1,272 @@
...
+
+const hasVideo = (sermon: Sermon) => {
+	const { videoStreams = [], videoFiles = [] } = sermon;
+
+	return videoStreams.length > 0 || videoFiles.length > 0;
+};
+
+interface SermonProps {
+	sermon: RecordingFragment;
+}
```
mattleff, line 49:
```suggestion
interface RecordingProps {
	recording: RecordingFragment;
}
```

src/components/organisms/recording.tsx
```diff
@@ -0,0 +1,272 @@
...
+import Player from '@components/molecules/player';
+import PlaylistButton from '@components/molecules/playlistButton';
+import SpeakerName from '@components/molecules/speakerName';
+import styles from '@containers/sermon/detail.module.scss';
+import { RecordingFragment } from '@lib/generated/graphql';
+import { readableBytes } from '@lib/readableBytes';
+import { makeSeriesRoute } from '@lib/routes';
+import useLanguageRoute from '@lib/useLanguageRoute';
+
+export type Sermon = NonNullable<RecordingFragment>;
```
mattleff, line 15: Can this be `Recording`?

src/lib/api/fetchApi.ts
```diff
@@ -10,17 +13,51 @@ export function storeRequest(request: IncomingMessage): void {
...
+	query = removeDuplicateFragments(query);
+
+	// TODO: Improve type
+	const headers: any = {
 		'Content-Type': 'application/json',
-		'x-av-session': cookies?.avSession,
 	};
 
+	if (cookies?.avSession) {
+		headers['x-av-session'] = cookies?.avSession;
```
mattleff, line 58: I don't think you need the `?` here.

src/lib/api/fetchApi.ts
```diff
@@ -10,17 +13,51 @@ export function storeRequest(request: IncomingMessage): void {
 	_request = request;
 }
 
+const removeDuplicateFragments = (query: string): string => {
+	const ast = parse(query);
+
+	const seen: string[] = [];
+	const newDefinitions: any[] = [];
+
+	ast.definitions.forEach((def) => {
```
mattleff, line 22: Instead of using `.forEach()` this would be clearer with a `.filter()`, IMO:

```ts
const newDefinitions = ast.definitions.filter((def) => {
    if (def.kind !== 'FragmentDefinition') {
        return true;
    }

    const id = `${def.name.value}-${def.typeCondition.name.value}`;
    const hasAlreadySeenFragment = seen.includes(id);
    seen.push(id);
    return !hasAlreadySeenFragment;
});
```

src/lib/api/fetchApi.ts
```diff
@@ -10,17 +13,51 @@ export function storeRequest(request: IncomingMessage): void {
 	_request = request;
 }
 
+const removeDuplicateFragments = (query: string): string => {
```
mattleff, line 16: This is a great solution for the bug. It would be good to include the link to the issue in a `// WORKAROUND: ...` comment so future maintainers would know if they still needed this.

src/lib/routes.ts
```diff
@@ -18,6 +18,14 @@ export const makeSeriesRoute = (
 	seriesId: string
 ): string => `/${languageRoute}/series/${seriesId}`;
 
+// TODO: url encode tagName
```
mattleff, line 21: It looks like you're already url encoding `tagName`...?

src/pages/[language]/sermons/all/page/[i].tsx
```diff
@@ -51,19 +39,39 @@ export async function getStaticProps({
 			description: 'All sermons feed title',
 		},
 		{
-			lang: lang?.display_name,
+			lang: display_name,
 		}
 	);
 
-	// TODO: Find a less-hacky way to generate feeds regularly.
```
Comment (mattleff) [inline src/pages/[language]/sermons/all/page/[i].tsx:58]: Do you still need this comment, or are you comfortable `getStaticProps()` will be called on a reasonable interval?

mattleff (approved):
Good work! :shipit:

src/pages/[language]/sermons/all/page/[i].tsx
Comment (narthur) [reply src/pages/[language]/sermons/all/page/[i].tsx:58]: I'm feeling ok with it. If the page is being regenerated, it seems ok to regenerate the xml file, too.

Comment (mattleff) [reply src/pages/[language]/sermons/all/page/[i].tsx:58]: Ok. That's up to every 10 seconds currently: https://github.com/avorg/audioverse-next/blob/27f723a7128d206718fd216c8987e3b58f1877c7/src/lib/getPaginatedStaticProps.ts#L67

src/components/organisms/recording.tsx
```diff
@@ -12,15 +12,18 @@ import { readableBytes } from '@lib/readableBytes';
 import { makeSeriesRoute } from '@lib/routes';
 import useLanguageRoute from '@lib/useLanguageRoute';
 
-export type Sermon = NonNullable<RecordingFragment>;
+export type RecordingType = NonNullable<RecordingFragment>;
```
mattleff, line 15: Do you need to export this type?

PR #39, opened 2021-02-02
add tags list page

narthur:


src/components/HOCs/withIntl.spec.tsx
```diff
@@ -36,7 +37,7 @@ type QueryByText = (
 ) => HTMLElement | null;
 
 const expectNoUnlocalizedText = (queryByText: QueryByText) => {
-	expect(queryByText(/[^z]+/)).not.toBeInTheDocument();
+	expect(queryByText(/[^z0-9]+/)).not.toBeInTheDocument();
```
mattleff, line 40: Would it be better to use a longer, more unique mock formatted message than `z`?

mattleff (approved):
Looks great! 🍓

src/components/HOCs/withIntl.spec.tsx
narthur: Hmm, in what way?

mattleff: Right now my brain reads this as "expect text other than `z` or numbers not to be in the document." That's less clear, IMO, than code that would read as "expect text other than `MOCK_LOCALIZED_MESSAGE` or numbers not to be in the document." Currently, it is not at all apparent why the absence of `z` answers/satisfies the function title of "expect no unlocalized text" (unless you already know that `z` is the mock formatted message).

PR #44, opened 2021-03-09
Sponsors

narthur:

PR #46, opened 2021-03-19
Add check TypeScript command

mattleff:
I noticed there were several TypeScript errors on #44, so I thought it would be handy to have a command to see those since it seems NextJS doesn't expose them.


narthur (approved):
This is great!

PR #47, opened 2021-04-12
Register

narthur:


src/lib/generated/graphql.ts
```diff
@@ -7878,7 +7942,7 @@ export const useAddPlaylistMutation = <
       options
     );
 import { fetchApi } from '@lib/api/fetchApi' 
-
+
```
mattleff, line 7945: This is from the extra space at line 33 in `graphql-plugin-getters.js`.

src/containers/account/register.tsx
```diff
@@ -0,0 +1,83 @@
...
+				placeholder={'email'}
+				required={true}
+				onChange={(e) => setEmail(e.target.value)}
+			/>
+			<input
+				type="password"
+				placeholder={'password'}
+				required={true}
+				onChange={(e) => {
+					setPass(e.target.value);
```
mattleff, line 46: "To brace, and not to brace (line 39, 53)...that is inconsistent..." 😜

src/containers/account/register.tsx
```diff
@@ -0,0 +1,83 @@
...
+	const { mutate, isLoading, data, isSuccess } = useRegisterMutation();
+
+	useEffect(() => {
+		if (data?.signup.errors.length) {
+			setErrors(data?.signup.errors.map((e) => e.message));
+		}
+	}, [data]);
+
+	if (isLoading) {
+		return <p>loading...</p>;
```
mattleff, line 20: i18n

src/containers/account/register.tsx
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	if (isSuccess && !errors.length) {
+		return <p>success</p>;
```
mattleff, line 24: i18n

src/containers/account/register.tsx
```diff
@@ -0,0 +1,83 @@
...
+			</ul>
+			<input
+				type="email"
+				placeholder={'email'}
+				required={true}
+				onChange={(e) => setEmail(e.target.value)}
+			/>
+			<input
+				type="password"
+				placeholder={'password'}
```
mattleff, line 43: i18n throughout this component

src/containers/account/register.tsx
```diff
@@ -0,0 +1,83 @@
...
+		return <p>loading...</p>;
+	}
+
+	if (isSuccess && !errors.length) {
+		return <p>success</p>;
+	}
+
+	return (
+		<>
+			{/* TODO: figure out a better key to use */}
```
mattleff, line 29: `e`?

src/containers/account/register.tsx
```diff
@@ -0,0 +1,83 @@
+import React, { useEffect, useState } from 'react';
+
+import { useRegisterMutation } from '@lib/generated/graphql';
+
+function Register(): JSX.Element {
+	const [errors, setErrors] = useState<string[]>([]);
+	const [pass, setPass] = useState<string>('');
```
mattleff, line 7: Somehow having `pass` and `pass2` makes it less obvious to me that this is `password` and not like some sort of 2-pass thing. Wdyt of renaming to something like `password` and `passwordConfirm`?

src/containers/account/register.spec.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+		userEvent.click(getByText('sign up'));
+
+		await waitFor(() => {
+			expect(getByText('success')).toBeInTheDocument();
+		});
+	});
+});
+
+// displays registration errors
```
mattleff, line 164: Looks like these tests were added. 👍🏻

src/containers/account/register.tsx
```diff
@@ -0,0 +1,83 @@
...
+						newErrors.push('email is required');
+					}
+					if (!pass || !pass2) {
+						newErrors.push('please type password twice');
+					}
+					if (pass !== pass2) {
+						newErrors.push('passwords do not match');
+					}
+					if (newErrors.length) {
+						setErrors(newErrors);
```
mattleff, line 69: Would it be cleaner to always call `setErrors(newErrors);` here (and not at line 57) since `newErrors` will be an empty array if there aren't any errors?

src/containers/account/register.tsx
```diff
@@ -0,0 +1,83 @@
...
+	if (isLoading) {
+		return <p>loading...</p>;
+	}
+
+	if (isSuccess && !errors.length) {
+		return <p>success</p>;
+	}
+
+	return (
+		<>
```
mattleff, line 28: Should this be a page template that could be styled later?

mattleff (approved):
Good work! 👍🏻

src/containers/account/register.tsx
narthur: I don't think so?

PR #52, opened 2021-04-30
Account settings

narthur:


src/containers/account/profile.tsx
```diff
@@ -1,25 +1,135 @@
...
-				value={givenName}
-				onChange={(e) => setGivenName(e.target.value)}
+		<form onSubmit={submit}>
+			<ul>
+				{errors.map((e) => (
+					<li key={e}>{e}</li>
+				))}
+			</ul>
+
+			<Input label={'email'} type={'email'} value={email} setValue={setEmail} />
```
mattleff, line 97: i18n on all the input labels

src/containers/account/profile.tsx
```diff
@@ -1,25 +1,135 @@
...
+
+	function submit(e: FormEvent<HTMLFormElement>) {
+		e.preventDefault();
+
+		const isPassUnconfirmed = !password !== !passwordConfirm;
+
+		if (isPassUnconfirmed) {
+			setErrors([
+				intl.formatMessage({
+					id: 'profile__unconfirmedPasswordError',
```
mattleff, line 56: This is probably overthinking, but should these ids be structured so that sorted alphabetically they are grouped? Like `profile__error*` (e.g., `profile__errorPasswordUnconfirmed`)? So that when translating you'd have all the errors together or maybe all the messages related to a field (like `profile__passwordLabel`, `profile__passwordErrorUnconfirmed`, etc.). Just a thought...

mattleff (approved):
Looks great! `<Input>` is nice!

src/containers/account/profile.tsx
narthur: Good thought!

PR #53, opened 2021-05-13
User Playlists

narthur:


src/containers/account/playlists.tsx
```diff
@@ -0,0 +1,83 @@
...
+			// TODO: Only invalidate `useGetAccountPlaylistsPageDataQuery`
+			await queryClient.invalidateQueries();
+		},
+	});
+
+	async function addPlaylist(e: FormEvent<HTMLFormElement>) {
+		e.preventDefault();
+
+		if (!title) {
+			setErrors(['missing title']);
```
mattleff, line 33: i18n here and below

src/containers/account/profile.tsx
```diff
@@ -1,25 +1,135 @@
...
-				value={givenName}
-				onChange={(e) => setGivenName(e.target.value)}
+		<form onSubmit={submit}>
+			<ul>
+				{errors.map((e) => (
+					<li key={e}>{e}</li>
+				))}
+			</ul>
+
+			<Input label={'email'} type={'email'} value={email} setValue={setEmail} />
```
mattleff, line 97: i18n for all these labels

src/containers/account/profile.tsx
```diff
@@ -1,25 +1,135 @@
...
-	// TODO: make sure givenName doesn't get populated with email on login (why??)
-	// TODO: fix err: "A component is changing an uncontrolled input of type text to be controlled."
 	return (
-		<form>
-			<input
-				type={'text'}
-				value={givenName}
-				onChange={(e) => setGivenName(e.target.value)}
+		<form onSubmit={submit}>
+			<ul>
```
mattleff, line 91: How about a `<FormErrors>` component that would take an array of errors and display it in a standard way?

mattleff (approved):
Looks great! 🍓

src/containers/account/profile.tsx
narthur: Punting until it's clearer what direction we're going for error handling.

PR #54, opened 2021-05-18
Header style

narthur:


mattleff (approved):
Looks amazing! 🤩  Two things I noticed:

1. The sidebar scrolls with long pages (such as http://localhost:3000/en/sermons/22275) and I don't think it should (but I might be wrong).
2. The icons look too small relative to their text vs. Figma.

PR #56, opened 2021-06-07
Home recent section cards

narthur:


src/containers/home.graphql
```diff
@@ -1,7 +1,50 @@
...
+		language: $language
+		tagName: "Family"
+		first: 1
+		orderBy: { field: PUBLISHED_AT, direction: DESC }
+	) {
+		nodes {
+			...cardTopic
+		}
+	}
+	recordings(
```
mattleff, line 41: Do you want to use `sermons` here since you've got `stories` and `musicTracks` above? You could end up with `recordings` returning the same recording as either the `stories` or `musicTracks` query.

src/components/molecules/cardStory.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+interface CardStoryProps {
+	story: CardStoryFragment;
+}
+
+export default function CardStory({ story }: CardStoryProps): JSX.Element {
+	const container = story.sequence
+		? {
+				icon: <FeatherIcon width={12} height={12} />,
+				title: story.sequence?.title,
```
mattleff, line 17: You shouldn't need a couple of these `?`.

src/components/molecules/cardSong.graphql
```diff
@@ -0,0 +1,18 @@
+fragment cardSong on Recording {
+	title
+	persons {
+		id
```
mattleff, line 4: Should this just be `...speakerName`?

src/components/molecules/cardSong.graphql
```diff
@@ -0,0 +1,18 @@
...
+		name
+		viewerHasFavorited
+		summary
+		website
+		imageWithFallback {
+			url(size: 100)
+		}
+	}
+	duration
+	collection {
```
mattleff, line 14: This should be `sequence`, right? I'm pretty sure no songs have collections.

src/containers/home.tsx
```diff
@@ -1,21 +1,237 @@
...
+						<h2>Recent content</h2>
+						<p>
+							Explore a select few of our audio pieces. Then when you’re ready,{' '}
+							<a href="#">create an account</a> to view even more.
+						</p>
+					</>
+				}
+				media={
+					<div className={styles.recent}>
+						{song && <CardSong song={song} />}
```
mattleff, line 89: Do we always want these in this order or should we sort them by the most recent first, or something?

mattleff (approved):
Great work! 👍🏻 🗼

src/containers/home.tsx
narthur: I'm going to say let's punt on this since I'm not certain how we'll be sourcing this content in the future.

PR #60, opened 2021-06-10
finish language switcher

narthur:


src/components/molecules/languageSwitcher.module.scss
```diff
@@ -0,0 +1,18 @@
+@import '../../styles/common';
+
+.languages {
+	text-decoration: underline;
```
mattleff, line 4: I think this needs a `text-underline-offset: 2px;` or so to match the Figma comp. I wonder if this link style should be extracted into something applying to general links as well as included here.

mattleff (approved):
Nice! Can we get rid of the ripple effect on the button? It looks terrible, IMO.

src/components/molecules/languageSwitcher.module.scss
narthur: Huh, I had no idea that property existed. TIL.

PR #61, opened 2021-06-10
Marketing header links

narthur:


src/containers/account/login.spec.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+		await renderPage({ router: { push } });
+
+		await waitFor(() => {
+			expect(push).toBeCalledWith('/en/discover');
+		});
+	});
+});
+
+// TODO:
```
mattleff, line 32: These have been done, right?

src/containers/home.tsx
```diff
@@ -27,29 +29,35 @@ export default function Home({ data }: HomeProps): JSX.Element {
...
+					>
 						<FormattedMessage
 							id={'homePage__downloadAppButtonLabel'}
 							defaultMessage={'Download App'}
 							description={'home page download app button label'}
 						/>
 					</a>
 					<span>|</span>
-					<a href={'#'}>
+					<a href={`/${route}/account/login`}>
```
mattleff, line 53: This should be conditional on auth state. If this link is only shown when the user is logged out then we shouldn't hit the immediate-redirect-to-discover on the login page.

src/containers/account/login.tsx
```diff
@@ -0,0 +1,17 @@
...
+import withAuthGuard from '@components/HOCs/withAuthGuard';
+import { useRouter } from 'next/router';
+import useLanguageRoute from '@lib/useLanguageRoute';
+
+function Login(): JSX.Element {
+	const router = useRouter();
+	const route = useLanguageRoute();
+
+	useEffect(() => {
+		router.push(`/${route}/discover`);
```
mattleff, line 12: I wonder if `account/login` should redirect at the next.js (server) level to `discover`. If discover is protected with `withAuthGuard` then the user would get the login form if they weren't logged in. Wdyt?

mattleff (approved):
Looks good. 👍🏻  I had just a few comments around handling the login page.

src/containers/home.tsx
narthur: Hmm. A lot of web apps I use do this on their marketing homes. Can you describe in more detail how you think it should work?

src/containers/account/login.tsx
narthur: I was assuming discover wouldn't be protected, since we want to technically support anonymous usage even if we don't promote it.

mattleff: Oh yeah, you're right... Hmm.

src/containers/home.tsx
mattleff: Yeah, nvm this. I was thinking discover would be protected and we'd switch this out for a Discover link when signed in. It always feels awkward to me to view a home page of a service I'm logged into and to not have any indication I'm logged into it, but I guess we can revisit this later if we get feedback from users.

PR #63, opened 2021-06-14
Recent posts

narthur:


src/containers/home.tsx
```diff
@@ -224,7 +226,10 @@ export default function Home({ data }: HomeProps): JSX.Element {
 								description={'home page recent posts section text'}
 							/>
 						</p>
-						<a href="#" className={`${styles.button} ${styles.primary}`}>
+						<a
+							href="https://www.audioverse.org/english/blog"
```
mattleff, line 230: Should this be a relative link?

src/containers/home.tsx
```diff
@@ -234,22 +239,11 @@ export default function Home({ data }: HomeProps): JSX.Element {
...
-							<FormattedMessage
-								id={'homePage__postTwoPlaceholder'}
-								defaultMessage={'post 2'}
-								description={'post two placeholder'}
-							/>
-						</div>
-					</>
+					<div className={styles.posts}>
+						{posts.map((p) => (
+							<CardPost key={p.title} post={p} />
```
mattleff, line 244: Should the key be `p.id`? Theoretically there could posts with duplicate titles.

src/types.d.ts
```diff
@@ -4,9 +4,11 @@ type ExactAlt<T, Shape> = T extends Shape
 		: never
 	: never;
 
-type PartialDeep<T> = {
-	[P in keyof T]?: PartialDeep<T[P]>;
-};
+// https://stackoverflow.com/a/61132308/937377
```
mattleff, line 7: Can this be deleted?

src/components/molecules/cardPlayable.tsx
```diff
@@ -0,0 +1,75 @@
...
+					? {
+							icon: container?.icon,
+							title: container?.title,
+					  }
+					: undefined
+			}
+			preTitle={
+				(container?.length &&
+					container?.index &&
+					`Part ${container?.index} of ${container?.length}`) ||
```
mattleff, line 47: i18n

src/components/molecules/cardPlayable.tsx
```diff
@@ -0,0 +1,75 @@
...
+			theme={theme}
+		>
+			<div className={styles.speakers}>
+				{persons.map((p) => (
+					<SpeakerName person={p} key={p.id} />
+				))}
+			</div>
+			<div className={styles.controls}>
+				{duration && (
+					<span className={styles.duration}>{formatDuration(duration)}</span>
```
mattleff, line 61: Do we need to handle i18n in `formatDuration()`?

src/components/molecules/cardPost.tsx
```diff
@@ -0,0 +1,28 @@
...
+			title={post.title}
+			url={post.canonicalUrl}
+			preTitle={d.toLocaleString('default', {
+				month: 'long',
+				day: 'numeric',
+				year: 'numeric',
+			})}
+		>
+			<p>{post.teaser}</p>
+			{post.readingDuration && <p>{formatDuration(post.readingDuration)}</p>}
```
mattleff, line 25: I wonder if we should have a minimum displayed reading duration. It looks weird to say `0m`:

image

src/components/molecules/cardPost.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { CardPostFragment } from '@lib/generated/graphql';
+
+interface CardPostProps {
+	post: CardPostFragment;
+}
+
+export default function CardPost({ post }: CardPostProps): JSX.Element {
+	const d = new Date(post.publishDate);
+	return (
+		<Card
```
mattleff, line 14: It feels like the whole card should be linked, not just the title. Or at least the hero image should also be linked.

src/components/molecules/cardPost.tsx
```diff
@@ -0,0 +1,28 @@
...
+import Card from '@components/molecules/card';
+import formatDuration from '@lib/formatDuration';
+import { CardPostFragment } from '@lib/generated/graphql';
+
+interface CardPostProps {
+	post: CardPostFragment;
+}
+
+export default function CardPost({ post }: CardPostProps): JSX.Element {
+	const d = new Date(post.publishDate);
```
mattleff, line 12: Do we want to display the publish date in the user's timezone? I guess I'm wondering about like someone in Hawaii viewing a `175 Years Ago Today` post dated `October 21, 2019` instead of `October 22, 2019` because the publish date was displayed relative to the user.

mattleff (approved):
LGTM! 🏣

src/containers/home.tsx
narthur: Added a todo.

src/containers/home.tsx
narthur: Given we're only showing the two most recent posts, and I'm not even sure anything would break if the keys were duplicates, I don't think it's worth fixing.

src/components/molecules/cardPost.tsx
narthur: I don't think we have a good way to fix this, since we'd need to know the local time the recording was published, and I think currently API published times are in UTC.

PR #65, opened 2021-06-15
Discover

narthur:


src/pages/_app.module.scss
```diff
@@ -12,5 +12,6 @@
 }
 
 .content {
-	flex: 1 0 100%;
+	flex: 0 0 calc(100% - 200px);
```
mattleff, line 15: I think this is messing up the marketing home page.

src/containers/discover.graphql
```diff
@@ -0,0 +1,10 @@
+query getDiscoverPageData($language: Language!) {
+	sermons(
```
mattleff, line 2: Re: [Trello card](https://trello.com/c/BcYH1pJA/231-does-it-make-sense-to-add-the-ability-to-query-the-api-for-recent-mixed-content): Should I build a basic query that just returns recent sermons for now, but would allow us to add sequences, collections, etc. later, or should we wait on the API until we actually want to implement it in the UI?

src/components/atoms/lineHeading.module.scss
```diff
@@ -0,0 +1,16 @@
+@import '../../styles/common';
+
+.heading {
+	color: $ts-red;
+	font-size: 16px;
+	text-transform: uppercase;
+	display: flex;
```
mattleff, line 7: I think we need `letter-spacing: 0.1em` in the heading, and `left: 2px` (or maybe 1?) in the `:after`.

src/containers/discover.tsx
```diff
@@ -1,5 +1,25 @@
...
+export default function discover({
+	data,
+}: {
+	data: GetDiscoverPageDataQuery;
+}): JSX.Element {
+	const sermons = data.sermons.nodes || [];
+	return (
+		<>
+			<LineHeading>Recent Content</LineHeading>
+			<Masonry
```
mattleff, line 17: There's a weird `:focus` `outline` style whenever you focus anywhere in the masonry grid.

mattleff (approved):
It's looking amazing! 🎨

src/containers/discover.graphql
narthur: @mattleff Let's wait until we're ready to add a new entity. That will save us both time while we're in crunch mode, and I don't think it will slow us down when we're ready to switch.

PR #82, opened 2021-09-13
Attempt to fix feed generation issues

mattleff:


src/lib/date.ts
```diff
@@ -1,7 +1,25 @@
 import dayjs from 'dayjs';
 
-export const formatLongDate = (date: string): string => {
+export const parseRelativeDate = (date: string): Date | undefined => {
```
narthur, line 3: I don't understand how this function is relative. It looks like it's just parsing a date string. Where is something relative happening?

src/pages/[language]/books/[id]/feed.xml/index.ts
```diff
@@ -0,0 +1,39 @@
...
+	const id = params?.id as string;
+	const languageRoute = params?.language as string;
+
+	const { audiobook: sequence } = await getAudiobookDetailPageData({
+		id,
+	}).catch(() => ({
+		audiobook: null,
+	}));
+
+	if (res && sequence) {
```
Comment (narthur) [inline src/pages/[language]/books/[id]/feed.xml/index.ts:23]: Should we add a todo for throwing a 404 on feeds if no data?

narthur (approved):
Great work!

src/pages/[language]/books/[id]/feed.xml/index.ts
Comment (mattleff) [reply src/pages/[language]/books/[id]/feed.xml/index.ts:23]: Ah, good point. I've got a fix locally and I'll send it with the next PR.

src/lib/date.ts
mattleff: Yeah, I'll add a comment for this with the next PR.

PR #84, opened 2021-09-15
First pass at bookmarking auth flow, refactoring

mattleff:
Fixes https://trello.com/c/LF1Ygzzs/86-make-bookmark-buttons-functional, https://trello.com/c/AjAdz4He/83-recording-share-menu-styling.

image

image


src/components/molecules/buttonShareRecording.tsx
```diff
@@ -29,6 +39,39 @@ export default function ButtonShareRecording({
...
+		[
+			twitterLink, // eslint-disable-next-line react/jsx-key
+			<FormattedMessage
+				id="molecule-buttonShareRecording__twitter"
+				defaultMessage="Twitter"
+			/>,
+		],
+		[
+			emailLink,
+			// eslint-disable-next-line react/jsx-key
```
narthur, line 59: Why are we disabling this rule instead of adding a key?

src/components/templates/andAuthBarrier.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+	const handleClose = () => {
+		setIsOpen(false);
+	};
+
+	const authBarrierContext: AuthBarrierContextType = {
+		challenge: () => {
+			setIsOpen(true);
+			throw new Error('Challenging for Authentication');
+		},
```
narthur, line 46: Nice!

narthur (approved):
Fantastic work!

src/components/molecules/buttonShareRecording.tsx
mattleff: I think it's a false warning, but it's probably better to have superfluous keys than disabling lint warnings: f22ab3d.

PR #125, opened 2021-11-22
Disable pointer events for non-interactive progress bar

mattleff:
Fixes https://trello.com/c/hjzukFVT/169-pointer-doesnt-indicate-progress-bar-in-card-is-clickable-but-clicking-opens-recording-confusing.


narthur (approved):
lgtm

PR #130, opened 2021-11-29
Disable bible card

narthur:
Ref: https://trello.com/c/NERf5GRw/170-pressing-play-on-alternating-cards-in-recent-content-doesnt-properly-switch-playing-media-then-pauses-unexpectedly


narthur:
> I think it would be better to actually fix the card rather than disable it. I think we could hard code the data on the home page (since we don't need an API request to get `Genesis 1`) and drop the play icon on the card. Does that make sense? Also, please run `npm ci` and `npm run graphql:codegen`. I think we could/should git ignore `dist/*`.

Sounds good: https://github.com/avorg/audioverse-next/pull/130/commits/7346754635efefe145ee49ffe3fbbf7cb7a48979

narthur:
@mattleff I'm still getting 404's locally for the bible book detail page... Does this branch break things somehow?

mattleff (approved):
This looks good. There's two things that need to get fixed before we ship this:

1. We should remove the `audiobible` query from `home.graphql` since we're not using it anymore.
2. We need to fix the Bible chapter card hat. I don't think we can ship it like this:

image

mattleff (approved):
LGTM! :shipit:

PR #131, opened 2021-11-30
Handle recording playback sessions

mattleff:
Fixes https://trello.com/c/jJqektrJ/97-recording-playback-sessions, https://trello.com/c/wEfrSXdq/178-what-do-we-think-about-making-v-h-gaps-between-cards-uniform-on-home-page, https://trello.com/c/3yxPgptc/181-fix-story-hat-colors, https://trello.com/c/GoDN2Hh4/180-make-sequence-detail-use-content-type-themes.


src/lib/time.ts
```diff
@@ -57,17 +57,29 @@ export const useFormattedDuration = (seconds: number): string => {
...
+					s: _.padStart(s.toString(), 2, '0'),
+				}
+		  )
+		: intl.formatMessage(
+				{
+					id: 'time__minutesAndSeconds',
+					defaultMessage: '{m}:{s}',
+					description: 'time minutes and seconds',
+				},
+				{ h, m, s: _.padStart(s.toString(), 2, '0') }
```
narthur, line 83: Is `h` needed here?

narthur (approved):
Great work!

src/lib/time.ts
mattleff: Good catch: 203e32e

PR #136, opened 2021-12-02
Add support for audio embed code

mattleff:
Fixes https://trello.com/c/zjjGoWaJ/36-media-embed-routes-https-wwwaudioverseorg-english-embed-media-18634.

Before and after:

image


src/components/organisms/miniplayer.tsx
```diff
@@ -5,7 +5,7 @@ import React, { useContext } from 'react';
 import ButtonNudge from '@components/molecules/buttonNudge';
 import ButtonPlay from '@components/molecules/buttonPlay';
 import RecordingProgressBar from '@components/molecules/recordingProgressBar';
-import { PlaybackContext } from '@components/templates/andMiniplayer';
+import { PlaybackContext } from '@components/templates/andPlaybackContext';
```
narthur, line 8: Nice extract!

src/components/molecules/player.spec.tsx
```diff
@@ -50,9 +51,11 @@ const recording: Partial<PlayerFragment> = {
 const renderComponent = buildRenderer(
 	(props: PlayerProps) => {
 		return (
-			<AndMiniplayer>
-				<Player {...props} />
-			</AndMiniplayer>
+			<AndPlaybackContext>
+				<AndMiniplayer>
```
narthur, line 55: What if `AndMiniplayer` used `AndPlaybackContext`? Is there any time we'd want to use `AndMiniplayer` without `AndPlaybackContext`?

src/containers/base.tsx
```diff
@@ -39,6 +40,12 @@ function Base<P>({
 }): JSX.Element {
 	const { description, disableSidebar, title, canonicalUrl, dehydratedState } =
 		pageProps;
+
+	useEffect(() => {
+		document.body.classList[disableSidebar ? 'add' : 'remove'](
+			'body--no-sidebar'
+		);
+	}, [disableSidebar]);
```
narthur, line 48: Could we do it [this way](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/toggle)?

```
document.body.classList.toggle('body--no-sidebar', disableSidebar)
```

src/containers/sermon/embed.module.scss
```diff
@@ -0,0 +1,92 @@
+@import '../../styles/common';
+
+.base {
+	// redefine these by content type in the future
```
narthur, line 4: Should this be a TODO?

src/containers/sermon/embed.module.scss
```diff
@@ -0,0 +1,92 @@
...
+	--recordingAccent: #{$ts-red};
+	border: 1px solid $ts-cream;
```
narthur, line 6: Nitpick, but would it be cleaner if this wasn't in the middle of all the custom prop definitions?

narthur (approved):
Great work!

src/components/molecules/player.spec.tsx
narthur: Maybe using an HOC pattern? 🤔

mattleff: The only reason we have to do this is because we're trying to test `<AndMiniplayer>` in isolation. Anywhere in the app `<AndPlaybackContext>` will be provided further up the tree (in `base.tsx`).

src/containers/base.tsx
mattleff: Good idea: 03f401c

src/containers/sermon/embed.module.scss
mattleff: Sure: 03f401c

src/containers/sermon/embed.module.scss
mattleff: I made it a `FUTURE: ` for now. I'm thinking before launch all the `TODO`s need to get converted either into `FUTURE:`s or actually fixed.

PR #141, opened 2021-12-10
Tweaks and fixes

mattleff:
Fixes a number of AV Redesign board issues.

image

image

### DIFF
--- a/src/containers/collection/detail.tsx
+++ b/src/containers/collection/detail.tsx
@@ -15,7 +15,6 @@ import CardRecording from '@components/molecules/card/recording';
 import CardSequence from '@components/molecules/card/sequence';
 import CardGroup from '@components/molecules/cardGroup';
 import CollectionTypeLockup from '@components/molecules/collectionTypeLockup';
-import ContentWidthLimiter from '@components/molecules/contentWidthLimiter';
 import DefinitionList, {
 	IDefinitionListTerm,
 } from '@components/molecules/definitionList';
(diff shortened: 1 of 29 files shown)


src/containers/sponsor/detail.graphql
```diff
@@ -9,7 +9,7 @@ query getSponsorDetailPageData($id: ID!) {
 		language
 		shareUrl
 		image {
-			url(size: 100)
```
narthur, line 12: I guess there's not a good way to extract this number into a constant?

src/containers/collection/detail.tsx
```diff
@@ -104,76 +103,83 @@ function CollectionDetail({
...
+							<Image
+								alt={title}
+								src={image.url}
+								height="500"
+								width="500"
+								objectFit="cover"
+							/>
+						</div>
+						{/* eslint-disable-next-line @next/next/no-img-element, @mizdra/layout-shift/require-size-attributes */}
+						<img alt={title} src={image.url} />
```
narthur, line 119: Why do we need to use an `` element here?

narthur (approved):
Great Work

src/containers/sponsor/detail.graphql
mattleff: Nope, not that I'm aware of unfortunately...

src/containers/collection/detail.tsx
mattleff: I need the image to scale with the actual size of the image generated (up to a max size). Using `next/image` doesn't seem to be possible since all the attributes are applied in the `style` attribute of the `img`.

narthur: @mattleff Would it make sense to extract this into its own component, something like "InherentSizeImage" or something, so that we can isolate the lint exception?

PR #148, opened 2021-12-20
Update langs and add extract new script

mattleff:


mattleff:
@narthur Feel free to review as you're able.

PR #204, opened 2022-01-18
Add redirects for legacy podcast routes

mattleff:
Fixes https://trello.com/c/68ko2Xcm/276-redirect-legacy-rss-urls.


narthur (approved):

src/lib/getLanguageIdByLegacyRoute.ts
```diff
@@ -0,0 +1,17 @@
...
+	SupportedLanguages,
+} from '@lib/constants';
+
+export function getLanguageIdByLegacyRoute(
+	route: string | undefined
+): SupportedLanguages | undefined {
+	return findKey(
+		LANGUAGES,
+		(l: LanguageConfiguration) =>
+			l.base_url === route || l.legacy_base_url === route
```
narthur, line 15: Is this function named correctly since it matches by both legacy and non-legacy routes?

narthur (approved):

src/lib/getLanguageIdByLegacyRoute.ts
mattleff: 42f8f11

mattleff:
I'm not sure why, but this is not working on the deploy preview....

PR #215, opened 2022-02-02
Attempt to fix Bible not playing on Android

mattleff:
Attempt to fix https://trello.com/c/ot17Ug1f/270-android-app-wont-play-the-bible-anymore. This replaces the default permanent status code (`308`) with `301` on the main language slug redirect. The `301` code is supported in basically any client in existence, vs. `308` only being supported in non-legacy clients (hopefully such as Android).


mattleff:
@narthur Can you test on your device after this is deployed?

PR #230, opened 2022-02-10
Delete account flow

mattleff:
Fixes https://trello.com/c/UrBJhrry/286-delete-account-button-on-profile-page-doesnt-seem-to-do-anything.

### DIFF
--- a/src/containers/account/profile.tsx
+++ b/src/containers/account/profile.tsx
@@ -1,3 +1,4 @@
+import { useRouter } from 'next/router';
 import React, { FormEvent, useEffect, useState } from 'react';
 import { FormattedMessage, useIntl } from 'react-intl';
 import { useQueryClient } from 'react-query';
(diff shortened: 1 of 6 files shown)


src/containers/account/profile.tsx
```diff
@@ -6,24 +7,55 @@ import Heading1 from '@components/atoms/heading1';
...
 	const [email, setEmail] = useState('');
 	const [password, setPassword] = useState('');
 	const [givenName, setGivenName] = useState('');
 	const [surname, setSurname] = useState('');
+	const [showingDeleteConfirm, setShowingDeleteConfirm] = useState(false);
+
+	const [acceptedDeletedImmediately, setAcceptedDeletedImmediately] =
+		useState(false);
+	const [acceptedLibraryLost, setAcceptedLibraryLost] = useState(false);
+	const [acceptedNoLogin, setAcceptedNoLogin] = useState(false);
```
narthur, line 41: I have this hunch that if you made the checkboxes required and used a form submit for the button, you could get away with leaving the checkboxes uncontrolled and not have to keep track of all this state...

mattleff: @narthur I'm also using those state vars to conditionally show the "Must acknowledge ..." statement and reset it when someone checks them all. I'm not sure I could do that (any easier) with required checkboxes.

PR #231, opened 2022-02-14
WIP: add freshdesk widget

narthur:
https://trello.com/c/IIhygpP8/207-help-widget

@mattleff This is ready for review but not merge, since I need to coordinate it with our switch to using our FreshDesk shared inbox.


mattleff (approved):
Works for me 👍🏻 The only thing I wondered about is that it will always obscure the Sound Doctrine tagline in the footer:

image

I don't know if there's a way around that or not...

narthur:
@mattleff Yes... and it also obscures the miniplayer on desktop. Trying to figure out a workaround... 😛

mattleff:
@narthur Can we close this since we're moving away from Freshdesk?

PR #237, opened 2022-02-16
Try formatting transcripts with newlines

mattleff:
Fixes https://trello.com/c/O9uHAKua/296-split-transcripts-per-sentence.


mattleff:
@narthur Having looked at a bunch of recordings locally I'm not sure this is a good idea. IMO, it ends up looking messier than one wall of text. If you think it's worth shipping here it is. 😀

narthur:
@mattleff Hmm. I still  think this is more readable than it was. What if we set a target word count per paragraph, like, say 150 or 200, and automatically grouped sentences to try to stay around that number? Just something real naive to avoid it looking so messy.

PR #239, opened 2022-02-16
Improve login error messages

mattleff:
Depends on https://github.com/avorg/audioverse-graphql/pull/526 and fixes https://trello.com/c/EObdnjba/300-let-user-know-if-email-does-not-exist.


narthur (approved):
Great work!

PR #248, opened 2022-02-24
Hide persons on sequence cards with recordings

mattleff:
This change matches the Figma designs.

Before:

image

After:

image


narthur (approved):
lgtm!

PR #256, opened 2022-03-17
Space action

leonboripgs:
Added space key press event catcher to pause and play the video on every screens.

### DIFF
--- a/src/lib/useGlobalSpaceDown.ts
+++ b/src/lib/useGlobalSpaceDown.ts
@@ -2,7 +2,7 @@ import { useEffect } from 'react';
 
 export default (callBack: () => unknown, disabled?: boolean) => {
 	const handleKeyDown = (event: KeyboardEvent) => {
-		if ((event.target as HTMLElement).tagName == 'INPUT' || event.key != ' ') {
+		if ((event.target as HTMLElement).tagName != 'BODY' || event.key != ' ') {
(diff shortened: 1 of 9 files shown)


narthur (changes requested):
Good work! However it looks like the current approach prevents the spacebar from being used to click selected items when using the keyboard to navigate, and also prevents the spacebar from being used in the search box.

leonboripgs:
> Good work! However it looks like the current approach prevents the spacebar from being used to click selected items when using the keyboard to navigate, and also prevents the spacebar from being used in the search box.

Fixed preventing from input in search box but not sure which case prevents to be used for selecting item.

src/lib/useGlobalSpaceDown.ts
```diff
@@ -0,0 +1,18 @@
+import { useEffect } from 'react';
+
+export default (callBack: () => unknown, disabled?: boolean) => {
+	const handleKeyDown = (event: KeyboardEvent) => {
+		if ((event.target as HTMLElement).tagName == 'INPUT' || event.key != ' ') {
```
mattleff, line 5: I wonder if this would be better as:

```suggestion
		if ((event.target as HTMLElement).tagName != 'BODY' || event.key != ' ') {
```

That would still allow spacebar selection of select menus, links, etc.

PR #258, opened 2022-03-25
Attempt to fix Bible versions list page 404ing

mattleff:
The default value of `revalidate` is `false ([ref](https://nextjs.org/docs/api-reference/data-fetching/get-static-props#revalidate)) which would mean that if any API were unavailable when the page was rendered that 404 would be cached indefinitely.

See also https://github.com/vercel/next.js/issues/25907#issuecomment-857267054

### DIFF
--- a/audit-ci.json
+++ b/audit-ci.json
@@ -1,7 +1,8 @@
 {
 	"low": true,
 	"allowlist": [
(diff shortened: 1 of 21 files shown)


narthur (approved):
Let's ship it!

PR #259, opened 2022-03-29
Update Jesse's pic

mattleff:


narthur (approved):
This image definitely shows his personality better

PR #268, opened 2022-04-11
Import videojs-chromcast and videojs-airplay

marcomiduri:
@mattleff It looks like we got some test errors to fix. Would you please explain in detail? How does the test work in the codebase?


src/components/templates/andPlaybackContext.tsx
```diff
@@ -399,6 +407,13 @@ export default function AndPlaybackContext({
 				preload: 'auto',
 				defaultVolume: 1,
 				sources,
+				techOrder: ['chromecast', 'html5'],
+				plugins: {
+					chromecast: {},
+					airPlay: {
+						addButtonToControlBar: false,
```
mattleff, line 414: I think this is correct because we want to use a custom button, but it doesn't look it's been added to this PR. Here's the designs for the buttons: https://www.figma.com/file/SKSajWfEkf56GQfZyI2S4a/Design-System?node-id=5073%3A258263 + https://www.figma.com/file/SKSajWfEkf56GQfZyI2S4a/Design-System?node-id=5093%3A256915.

src/components/templates/andPlaybackContext.tsx
```diff
@@ -167,7 +167,15 @@ export default function AndPlaybackContext({
 
 	const [videojs, setVideojs] = useState<typeof VideoJs>();
 	useEffect(() => {
-		import('video.js').then((v) => setVideojs(v));
+		import('video.js').then((v) => {
+			// eslint-disable-next-line @typescript-eslint/no-var-requires
+			require('@silvermine/videojs-airplay')(v.default);
```
mattleff, line 172: Similarly to #269 this should use `import()` instead of `require()` to avoid bloating the global bundle.

src/styles/styles.scss
```diff
@@ -5,6 +5,8 @@
 // START normalize.css
 
 @import '../../node_modules/normalize.css/normalize.css';
+@import '../../node_modules/@silvermine/videojs-airplay/dist/silvermine-videojs-airplay.css';
```
mattleff, line 8: Are these needed if we're using custom buttons?

mattleff (changes requested):

mattleff:
@marcomiduri Re: test failures - We have a function that mocks Video.js when in test and you may need to update that mock to work with the AirPlay and Chromecast plugins: https://github.com/avorg/audioverse-next/blob/master/src/lib/test/helpers.tsx#L261

src/components/templates/andPlaybackContext.tsx
leonboripgs: * Just from my understanding import `video.js` should be done from the global side rather than at component level. Currently importing will be duplicated which is not the best practice.
* Instead of var requires (force terminated the eslint check), need to use ES6 style import.

src/components/templates/andPlaybackContext.tsx
```diff
@@ -412,6 +427,12 @@ export default function AndPlaybackContext({
 				resetPlayer();
 			} else {
 				import('video.js').then((videoJsImport) => {
+					// eslint-disable-next-line @typescript-eslint/no-var-requires
+					require('@silvermine/videojs-airplay')(videoJsImport);
```
leonboripgs, line 431: The same thing applies to here.

src/styles/styles.scss
leonboripgs: This style is not only for the buttons guess there are much more. Btw, from the documentation seems like better way to import styles and override variables which seems better than this.
`@import "path/to/node_modules/@silvermine/videojs-airplay/src/scss/videojs-airplay";`
And here - https://github.com/silvermine/videojs-airplay#using-the-css-and-images 
Maybe need to add the default buttons and customize the styles using this section.

src/components/templates/andPlaybackContext.tsx
leonboripgs: We are going to add custom buttons to here, but with difference appearance. What about `addButtonToControllBar: true` (by default) and customize it using the styles?

https://github.com/silvermine/videojs-airplay#using-the-css-and-images

leonboripgs (commented):
Just left few comments on the PR, once I get reply for them, I will approve it. Thank you.

src/components/templates/andPlaybackContext.tsx
marcomiduri: @mattleff I have fixed this issue

PR #272, opened 2022-04-11
Use website featured collection query

mattleff:
Depends on https://github.com/avorg/audioverse-graphql/pull/547.


narthur (approved):
Looking good!

PR #276, opened 2022-04-13
Don't offer opting out of newsletter on signup

mattleff:


narthur (approved):
Looks good. Do we have it on the roadmap to add this back in later?

PR #282, opened 2022-04-25
Fallback to user session token from headers

mattleff:
This allows the mobile app sending the session token as a header instead of as a cookie. Refs .


narthur (approved):
lgtm!

PR #225, opened 2022-02-10
Fix Safari two click for play bug

mattleff:
Fixes https://trello.com/c/MpvFmwTD/295-two-clicks-required-to-play-in-safari. This needs thorough testing because there's a lot of changes...

PR #285, opened 2022-05-11
Refactor test helpers and related types

narthur:


.eslintrc.js
```diff
@@ -31,6 +32,28 @@ module.exports = {
 		jest: true,
 	},
 	rules: {
+		'@calm/react-intl/missing-formatted-message': [
+			2,
```
mattleff, line 36: Should this (and below) be `error` or `warn`?

src/components/templates/andPlaybackContext.tsx
```diff
@@ -165,10 +165,7 @@ export default function AndPlaybackContext({
 	const videoElRef = useRef<HTMLVideoElement>(null);
 	const originRef = useRef<HTMLDivElement>(null);
 
-	const [videojs, setVideojs] = useState<typeof VideoJs>();
-	useEffect(() => {
```
mattleff, line 169: Don't we still need this?

src/lib/getIntlMessages.ts
```diff
@@ -4,8 +4,8 @@ import getLanguageByBaseUrl from '@lib/getLanguageByBaseUrl';
 
 export default function getIntlMessages(
 	languageRoute: string
-): Promise<ResolvedIntlConfig['messages']> {
+): ResolvedIntlConfig['messages'] {
 	const lang = getLanguageByBaseUrl(languageRoute, 'en');
 
-	return import(`../../public/compiled-lang/${lang?.base_url}.json`);
+	return lang?.messages || {};
```
mattleff, line 10: I think not using `import()` here will increase the bundle size.

src/lib/getNavigationItems.tsx
```diff
@@ -394,6 +395,16 @@ export function getNavigationItems(
 				description: `Header nav link name: More`,
 			}),
 			children: [
+				{
```
mattleff, line 398: AFAIK, this is for HelpScout.

tsconfig.json
```diff
@@ -26,7 +26,8 @@
 			"@lib/*": ["src/lib/*"],
 			"@components/*": ["src/components/*"],
 			"@containers/*": ["src/containers/*"],
-			"@pages/*": ["src/pages/*"]
+			"@pages/*": ["src/pages/*"],
+			"@root/*": ["*"]
```
mattleff, line 30: Is this change needed?

mattleff (changes requested):
Looks good overall! It seems the checks have stopped running and I think they would catch a few issues I noted in the comments. I also think the additional langs need a `npm run extract && npm run compile`.

mattleff:
@narthur The changes in this PR have significant bundle implications. I've highlighted a couple of them in my comments.

src/components/templates/andPlaybackContext.tsx
```diff
@@ -400,20 +397,17 @@ export default function AndPlaybackContext({
 				defaultVolume: 1,
 				sources,
 			};
+
 			if (playerRef.current) {
 				playerRef.current.src(sources);
 				resetPlayer();
-			} else if (videojs) {
```
mattleff, line 406: I believe this reintroduces the bug #225 fixed.

tsconfig.json
narthur: Good catch, resolved!

narthur:
@mattleff I just ran `npm run extract && npm run compile` and it didn't result in any changes... 😕

src/components/templates/andPlaybackContext.tsx
mattleff: @narthur I can reproduce the bug in Safari: 

1. In a Safari incognito window (or anywhere with uncached JS) open the site.
2. Click play on one of the home page recording cards.

### Expected result

The recording plays. This is the current behavior on the public website.

### Actual result

The recording loads the miniplayer but doesn't start playing.

mattleff (changes requested):
Looks great, except for the Safari two-click bug

mattleff (approved):
Can't reproduce the error anymore... Let's ship it and see if we get complaints.

PR #286, opened 2022-05-16
Json ld and next-pwa tuning. 

bsalb:
### Description
  - Here i have exclude the languages folder present inside public directory from being pre-cached. And added few more JSON-ld scripts for discover and similar pages with type Search box. I have also remove the aggregate rating from home page and localize the Audio verse slogan.

# To do:

- [x] Exclude the languages from being pre-cached.
- [x] Added localization in Home Page OG tags.
- [x] Added more JSON-ld scripts
- [x] Slightly improved initial bandwidth usage.


narthur:
@bsalb Apply the `ready-for-internal` label and then re-request reviews from Matthew and me when this is ready for review.

src/components/organisms/recording.tsx
```diff
@@ -249,6 +250,13 @@ export function Recording({
 
 	return (
 		<Tease className={clsx(styles.base, styles[contentType])}>
+			<CourseJsonLd
```
narthur, line 253: Not sure what the best place is to find info on JsonLd, but from this link I think course may be the wrong choice here:

https://jsonld.com/json-ld-course/

src/containers/home.tsx
```diff
@@ -78,8 +79,23 @@ export default function Home({ data }: HomeProps): JSX.Element {
...
+
 	return (
 		<>
+			<BrandJsonLd
+				slogan={formatedSlogan}
+				id="https://www.audioverse.org"
+				logo="/favicon.svg"
+				// aggregateRating={{
+				// 	ratingValue: '5',
+				// }}
```
narthur, line 97: I think we can delete `aggregateRating` entirely rather than commenting it out.

src/containers/bible/versions.tsx
```diff
@@ -23,6 +24,15 @@ export interface VersionsProps {
 function Versions({ versions }: VersionsProps): JSX.Element {
 	return (
 		<Tease className={styles.tease} fullBleed={false}>
+			<SiteLinksSearchBoxJsonLd
```
narthur, line 27: Still new to json ld, but should there just be one of these in home.tsx or something with all the actions in one place? Or is there an advantage to specifying the actions in different files?

narthur: Also, why are we defining the actions as searches, rather than directly linking to the page?

narthur (changes requested):
Great work! Looks like you also need to fix the test suite.

src/containers/home.tsx
bsalb: yes we can do that

src/containers/bible/versions.tsx
bsalb: we have some unique json-ld type on each of pages. So, i provide these search types on them. Regarding the actions, i found search box common. so i provide query_input as per page. For example: in bible page search box searches for bibles.

PR #294, opened 2022-06-02
Helpscout

narthur:


narthur:
@mattleff I think the idea behind not having a backdrop is to allow the user to use the site while referring to a help article.

.eslintrc.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
 		project: 'tsconfig.json',
 		sourceType: 'module',
 	},
-	ignorePatterns: ['**/*.js'],
+	ignorePatterns: ['**/*.js', 'src/lib/generated/**/*'],
```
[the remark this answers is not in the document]
narthur: I was trying to speed up the eslint run. Do you think we need to run eslint on graphql.ts?

src/lib/useNavigationItems.tsx
```diff
@@ -24,66 +25,85 @@ import {
...
 	children?: INavigationItem[];
 	childNode?: JSX.Element;
 };
 
-export function getNavigationItems(
-	router: NextRouter,
-	intl: IntlShape,
-	languageRoute: string
-): INavigationItem[] {
+// TODO: Find a way to reduce boilerplate without increasing bundle size
```
mattleff, line 42: Do we need this comment?

src/styles/mixins.scss
```diff
@@ -5,3 +5,40 @@
 		font-family: 'Inter var', 'system-ui', sans-serif;
 	}
 }
+
+// ********************
+// fadeIn
+// ********************
+// https://gist.github.com/chasewoodford/2939636
+$prefix: '-moz-', '-webkit-', '-o-', '-ms-', '';
```
mattleff, line 13: I don't think we need these prefixes since Next uses autoprefixer automatically: https://nextjs.org/docs/advanced-features/customizing-postcss-config#default-behavior

mattleff (changes requested):
LGTM! The lockfile is in version 1, once that's fixed let's ship it. 👍🏻

mattleff (approved):
👍🏻 :shipit:

PR #296, opened 2022-06-08
fix the beacon

narthur:


src/styles/styles.scss
```diff
@@ -201,7 +201,7 @@ $close-button-height: 30px;
 $beacon-height: $button-height + $close-button-height + $grid-gap * 3;
 
 .BeaconContainer {
-	@media (min-width: $breakpoint-lg) {
+	@media (min-width: 451px) {
```
mattleff, line 204: Can we pull this out to a variable to avoid magic number?

public/helpscout.js
```diff
@@ -1 +1,12 @@
-!function(e,t,n){function a(){var e=t.getElementsByTagName("script")[0],n=t.createElement("script");n.type="text/javascript",n.async=!0,n.src="https://beacon-v2.helpscout.net",e.parentNode.insertBefore(n,e)}if(e.Beacon=n=function(t,n,a){e.Beacon.readyQueue.push({method:t,options:n,data:a})},n.readyQueue=[],"complete"===t.readyState)return a();e.attachEvent?e.attachEvent("onload",a):e.addEventListener("load",a,!1)}(window,document,window.Beacon||function(){});
+!function(e, t, n) {
```
mattleff, line 1: Can we revert this back to the minimized version?

mattleff (approved):
Two comments, otherwise LGTM!

PR #297, opened 2022-06-08
Update MUI

narthur:


mattleff (changes requested):
It looks like there are a few visual regressions in the miniplayer:

image

PR #300, opened 2022-06-09
Replace React FB login dependency

mattleff:
Attempt to fix CI peer dependencies errors.


narthur:
@mattleff Since it doesn't look like this is working yet, I've removed myself from reviewers. Feel free to re-add me when you're ready.

PR #302, opened 2022-06-10
Eslint perf

narthur:


.eslintrc.js
```diff
@@ -4,7 +4,6 @@ module.exports = {
 		project: 'tsconfig.json',
 		sourceType: 'module',
 	},
-	ignorePatterns: ['**/*.js', 'src/lib/generated/**/*'],
 	plugins: [
 		'@typescript-eslint/eslint-plugin',
 		'import',
```
mattleff, line 9: If we're going to turn off import-order we should remove the `import` plugin (`eslint-plugin-import` in package.json).

.github/workflows/ci.yml
```diff
@@ -37,27 +37,71 @@ jobs:
         run: npm ci
       - name: Lint code
         run: npm run lint
-  build:
+        timeout-minutes: 1
+  prettier:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Setup Node.js
```
mattleff, line 44: We still need to setup node here.

mattleff (approved):
👍🏻 Looks good

.github/workflows/ci.yml
narthur: Nope. Prettier doesn't need it. Unless I'm missing something?

mattleff: ok, that's fine.

PR #309, opened 2022-06-23
add custom pages route

narthur:
This allows us to create custom pages through the admin dashboard. I decided to add it because we need a custom landing page for when users complete the opt-in process for the newsletter.

### DIFF
--- a/src/components/molecules/navItem.tsx
+++ b/src/components/molecules/navItem.tsx
@@ -18,7 +18,7 @@ export default function NavItem({
 
 	const inner = (
 		<>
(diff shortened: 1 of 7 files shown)


src/components/molecules/navItem.tsx
```diff
@@ -18,7 +18,7 @@ export default function NavItem({
 
 	const inner = (
 		<>
-			<span className={styles.icon}>{Icon && <Icon />}</span>
+			<div className={styles.icon}>{Icon && <Icon />}</div>
```
narthur, line 21: This is to avoid a nextjs hydration error.

PR #335, opened 2022-11-14
trigger beacon suggestions on all route changes

narthur:
This is a fix for our integration with HelpScout.


jlaverde77 (approved):
Looks good.

PR #337, opened 2022-11-16
Upgrade to NextJS 13

narthur:
https://trello.com/c/Ao5aiU8R/432-upgrade-nextjs-to-v13


jlaverde77 (commented):
This is a VERY large PR. Are we sure we don't want this broken up? Say, making the language changes into its own PR, the upgrade its own PR, etc?

jlaverde77 (approved):

PR #349, opened 2022-12-28
Init gitstream

narthur:
Add base config files for GitStream to enable features like automatically labeling PRs with how much time they'll take to review

https://linearb.io/dev/gitstream/


jlaverde77 (approved):
Talked about this change with Nathan. This is a dev tool which doesn't affect prod.

PR #354, opened 2023-01-05
add e2e workflow

narthur:
Previously, the Lighthouse check triggered immediately and then waited until the Vercel build succeeded before checking the pages. This could result in timeouts if multiple Vercel builds were queued.

This PR creates a separate E2E workflow that is only triggered once the Vercel deployment has succeeded, thereby making it unnecessary for the Lighthouse check to do any waiting itself.


jlaverde77 (approved):
Talked about this with Nathan. Looks very useful as it fixes broken pipeline.

PR #356, opened 2023-01-06
CODEOWNERS updates

narthur:
This PR:

- Removes Matthew from CODEOWNERS
- Configures CODEOWNERS to not request a review from Jose for Dependabot PRs

GitHub indicates that the modified file is valid:

<https://github.com/avorg/audioverse-next/blob/dep-codeowners/.github/CODEOWNERS>


jlaverde77 (approved):
Per @narthur this is only for notification, not requirement or reviewer, which makes sense.

PR #357, opened 2023-01-06
Graphql codegen optimization

narthur:
This is a small refactor to `graphql-plugin-react-query` to both make the plugin more readable, and remove unneeded whitespace from the plugin's output.


jlaverde77 (approved):
Nathan explained that this gets rid of whitespace from the function literal, so he created the function and referenced it directly.

PR #359, opened 2023-01-08
Lighthouse comments

narthur:
This PR modifies the Lighthouse E2E CI workflow to prevent it from creating multiple comments. Instead, on each run, it will update the existing comment if one exists.


jlaverde77 (approved):
GitHub Actions in workflows change to reduce redundant comments.

PR #363, opened 2023-02-09
AV-98 Removed amazon smile blurb from donate page

avorg:
Removed amazon smile blurb from donate page and horizontal rule.


narthur (approved):
Looks good! Can you also run `npm run extract` and `npm run compile` since this change removes localized strings?

PR #401, opened 2023-05-12
AV-48: Convert Discover page sections to use carousel-like pagination

narthur:
Story: https://audioverse7.atlassian.net/jira/software/projects/AV/boards/3?selectedIssue=AV-374


jlaverde77 (commented):
Carousel stopped working after I signed in. Also, it took a few seconds to load the carousel. Not sure if this is only because it's in preview, but if it's not, this will hurt UX.

narthur:
@jlaverde77 Good catch! Looks like I have some bugs to track down. 👍

PR #436, opened 2023-06-28
Disable react scope rule

narthur:
This PR disables the `react/react-in-jsx-scope` rule. This rule is no longer needed when using a newer version of the JSX transform. This PR also refactors our React imports to remove the unneeded default import.

Reference:

- https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/react-in-jsx-scope.md
- https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#removing-unused-react-imports

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -24,6 +24,7 @@ module.exports = {
 		'plugin:testing-library/dom',
 		'plugin:jest-dom/recommended',
 		'plugin:@next/next/recommended',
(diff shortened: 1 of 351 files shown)


narthur:
We're using a custom Babel config, so we can't actually do this without manually updating our build configuration to use the new JSX transform.

PR #455, opened 2023-08-08
AV-448

narthur:
Card:

[when full-matching occurs for any "Content Type", display that "Content Type" at the top of Search Results (website only)](https://audioverse7.atlassian.net/jira/software/projects/AV/boards/3?selectedIssue=AV-448)


jlaverde77 (approved):
@jperegrino75  already did functionality testing here. Code changes look good.

PR #470, opened 2023-09-19
Av 498 prompt users with warning message when clicking on continue as guest website and mobile browser

okeinom:


package.json
```diff
@@ -63,6 +63,7 @@
         "sass": "^1.42.1",
         "striptags": "^3.2.0",
         "swiper": "^9.4.1",
+        "ts-node": "^10.9.1",
```
narthur, line 66: Why did we add `ts-node` as an explicit dependency? It looks like Jest already includes that as a dependency:

```
work:~$ npm ls ts-node
audioverse-next@0.1.0 /Users/work/ProgrammingProjects/audioverse-next
├─┬ jest@27.5.1
│ └─┬ @jest/core@27.5.1
│   └─┬ jest-config@27.5.1
│     └── ts-node@10.9.1 deduped
└── ts-node@10.9.1
```

src/components/molecules/buttonGuest.tsx
```diff
@@ -29,14 +29,20 @@ export default function ButtonGuest({
...
+					onClick={(e) => {
+						e.preventDefault();
+						setIsGuestModalOpen(true);
+					}}
+				>
+					<FormattedMessage
+						id="molecule-buttonGuest__label"
+						defaultMessage="Continue as guest"
+					/>
+				</button>
```
narthur, line 44: Could we make this all one button since I think both buttons in this div do the same thing?

src/components/molecules/buttonGuest.tsx
```diff
@@ -29,14 +29,20 @@ export default function ButtonGuest({
...
-					<a className="decorated">
-						<FormattedMessage
-							id="molecule-buttonGuest__label"
-							defaultMessage="Continue as guest"
-						/>
-					</a>
-				</Link>
+				<button
+					className={styles.link}
+					data-testid="guest-info-button2"
```
narthur, line 34: We should avoid using `data-testid` whenever possible.

src/containers/account/register.spec.tsx
```diff
@@ -306,6 +306,14 @@ describe('register page', () => {
 
 		expect(getByText('Continue as guest?')).toBeInTheDocument();
 	});
+
+	it('pops modal on guest info click', async () => {
+		const { getByText, getByTestId } = await renderPage();
+
+		await userEvent.click(getByTestId('guest-info-button2'));
```
narthur, line 313: Let's use something like the following instead of `getByTestId`:

```
getByRole('button', {
  name: /Continue as guest/
})
```

src/containers/sermon/detail.spec.tsx
```diff
@@ -399,12 +399,12 @@ describe('sermon detail page', () => {
 					},
 				},
 			],
-			recordingDate: '253-03-01 09:30:00',
+			recordingDate: '2023-03-01 09:30:00',
```
narthur, line 402: Good catch!

narthur (dismissed):
Good work! Message me in Slack if you have any questions.

narthur:
@okeinom It also looks like you need to run `npm run intl:extract && npm run intl:compile`

package.json
okeinom: My Bad, it was saying something like it could not find it "ts-node"

src/components/molecules/buttonGuest.tsx
okeinom: I believe we could.

src/components/molecules/buttonGuest.tsx
okeinom: ok. I need to understand this test program a bit more, was just following what I saw. :)

src/containers/account/register.spec.tsx
okeinom: kk, will test run this.

jlaverde77 (approved):

PR #479, opened 2023-09-26
Av 518 presenters with no images to display first letter of first and last name presenter s card on website and mobile browser

okeinom:
I have updated person.tsx and added a component to create the initial circle


jlaverde77 (approved):
Looks good!

PR #523, opened 2024-03-19
updated play tracking to track content type

okeinom:


src/components/templates/andMiniplayer.graphql
```diff
@@ -40,13 +40,7 @@ query getRecordingExtraData($id: ID!) {
...
 			name
 		}
-		collection {
-			title
-		}
-
-		sequence {
-			title
-		}
+		contentType
```
alangumer, line 43: All the properties in this query but publishDate are already available in top components and can be passed down. As for publishDate, you can add it to a top query/fragment that makes sense, and pass it down too.

src/lib/usePlaybackSession.tsx
```diff
@@ -107,10 +107,11 @@ export default function usePlaybackSession(
 				Video: context.isShowingVideo(),
 			},
 			sponsor_title: trackData.data?.recording?.sponsor?.title,
-			series_title: trackData.data?.recording?.sequence?.title,
-			conference_title: trackData.data?.recording?.collection?.title,
+			series_title: recording?.sequence?.title,
+			conference_title: recording?.collection?.title,
 			Speakers: trackData.data?.recording?.speakers,
```
alangumer, line 112: All the keys should be lowercase.

alangumer (dismissed):

src/lib/usePlaybackSession.tsx
okeinom: @alangumer , I tried to update the files before doing this way. When I touch the the other files, tests failed and ran out of time.

avorg (approved):

PR #545, opened 2024-06-20
fix: private playlist

alangumer:
https://audioverse7.atlassian.net/browse/AV-1163


okeinom:
@alangumer  Looks good and straight forward.

PR #546, opened 2024-06-24
Prevent video from pausing when user navigates away from associated detail page

narthur:
Apologies for the super ugly hack. We spent a lot of time over here trying to find the cause of the bug so we could fix it properly, but we failed in the end.


src/components/templates/andPlaybackContext.tsx
```diff
@@ -533,6 +535,14 @@ export default function AndPlaybackContext({
 		}
 
 		destination.appendChild(video);
+
+		// WORKAROUND: player pauses when moving to miniplayer
+		if (_paused) {
```
alangumer, line 540: Use playback.paused() direclty.

alangumer (changes requested):

src/components/templates/andPlaybackContext.tsx
narthur: @alangumer I think this would defeat the purpose, since the goal is to restore the player to the state it was before the move. If I checked what state it was after the move, I wouldn't be restoring to what it was before the move.

alangumer: @narthur I see, ok.

alangumer (approved):

PR #549, opened 2024-06-25
Add Share Link

jperegrino75:
Added "Sharing Tools" Link to more menu


src/lib/useNavigationItems.tsx
```diff
@@ -470,6 +470,16 @@ export function useNavigationItems(): INavigationItem[] {
 						defaultMessage: 'Get Help',
 					}),
 				},
+				{
+					key: 'sharingtools',
+					href: 'http://share.audioverse.org/',
```
alangumer, line 475: Should be https and there is no need for the slash at the end.

alangumer (changes requested):
The translations are missing.

src/lib/useNavigationItems.tsx
jperegrino75: done and committed

alangumer (approved):

jperegrino75:
> The translations are missing.

we will keep the link to English in all languages as this is the way the other pages are currently.

jperegrino75:
@alangumer I resolved the conflicts by removing the code. This was for internationalization of the Sharing Link but since this is the same for all languages; I just removed the code. Hope I did it in the right way :-)

PR #564, opened 2024-08-07
Continue Listening Section

narthur:
This PR is for the purpose of completing [AV-78](https://audioverse7.atlassian.net/browse/AV-78?atlOrigin=eyJpIjoiNDAyZDhiNTA2NTBmNDljZmEzNTc4YzFjZjQwN2E5YjIiLCJwIjoiaiJ9).

- [x] when user is signed in, display “Continue Listening” section using the same API call of the APP
- [x] when guest, display “Featured Teachings”
- [x] Move featured section to top of discover page
- [x] Hide featured section for logged in users

[AV-78]: https://audioverse7.atlassian.net/browse/AV-78?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/organisms/cardSlider/section/index.module.scss
+++ b/src/components/organisms/cardSlider/section/index.module.scss
@@ -19,8 +19,3 @@
 		text-decoration: none;
 	}
 }
-
-.hidden {
-	height: 0px;
-	overflow: hidden;
-}
--- a/src/components/organisms/cardSlider/section/index.module.scss
+++ b/src/components/organisms/cardSlider/section/index.module.scss
@@ -19,3 +19,8 @@
 		text-decoration: none;
 	}
 }
+
+.hidden {
+	height: 0px;
+	overflow: hidden;
+}
--- a/src/components/organisms/cardSlider/section/index.tsx
+++ b/src/components/organisms/cardSlider/section/index.tsx
@@ -64,7 +64,7 @@ export default function Section<T extends GraphqlInfiniteQuery, N>({
 }: SectionProps<T, N>): JSX.Element {
 	const language = useLanguageId();
 
-	const { data, fetchNextPage } = infiniteQuery(
+	const { data, fetchNextPage, isFetchingNextPage } = infiniteQuery(
 		'after',
 		{ language },
 		{
@@ -90,11 +90,11 @@ export default function Section<T extends GraphqlInfiniteQuery, N>({
 
 	const preload = useCallback(
 		({ index, total }: { index: number; total: number }) => {
-			if (index + PRELOAD_COUNT >= total) {
+			if (index + PRELOAD_COUNT >= total && !isFetchingNextPage) {
 				fetchNextPage();
 			}
 		},
-		[fetchNextPage]
+		[fetchNextPage, isFetchingNextPage]
 	);
 
 	// Check if there's content to render, if not, return an empty JSX element
--- a/src/components/organisms/cardSlider/section/index.tsx
+++ b/src/components/organisms/cardSlider/section/index.tsx
@@ -115,9 +115,7 @@ export default function Section<T extends GraphqlInfiniteQuery, N>({
 	);
 
 	// Check if there's content to render, if not, return an empty JSX element
-	const renderFail = cards.length < 1;
-
-	if (renderFail) {
+	if (cards.length < 1) {
 		return <></>;
 	}
 
--- a/src/components/organisms/cardSlider/section/index.tsx
+++ b/src/components/organisms/cardSlider/section/index.tsx
@@ -83,10 +83,7 @@ export default function Section<T extends GraphqlInfiniteQuery, N>({
 				getNextPageParam: (last: Maybe<GraphqlInfiniteQuery>) => {
 					if (!last) return;
 					const pageInfo = selectPageInfo(last);
-					if (!pageInfo?.hasNextPage) {
-						console.warn('No more pages to fetch');
-						return;
-					}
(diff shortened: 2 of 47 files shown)


src/containers/discover/index.tsx
```diff
@@ -9,14 +10,28 @@ import StorySeasons from '~components/organisms/cardSlider/section/storySeasons'
...
+	const sessionToken = getSessionToken();
+
+	const authResult = useGetWithAuthGuardDataQuery(
+		{},
+		{
+			enabled: !!sessionToken,
+			retry: false,
+		}
+	);
+	const user = authResult.data?.me?.user;
```
narthur, line 30: Let's extract this into a new hook in lib/hooks, something like useMe.ts, and give it its own graphql query, useMe.graphql. And let's use the same hook inside withAuthGuard.

narthur (commented):
Looking good!

narthur (commented):
lgtm! 🚀

src/components/organisms/cardSlider/section/index.tsx
```diff
@@ -112,7 +155,11 @@ export default function Section<T extends GraphqlInfiniteQuery, N>({
 					</a>
 				)}
 			</LineHeading>
-			<CardSlider {...props} onIndexChange={preload} items={cards} />
+			{isLoading && showLoading ? (
+				<CardSlider {...props} items={loadingCards} />
```
narthur, line 159: Now that we're server-rendering the discover page, can we rip out the loading state we added? If I'm thinking right, it would never be visible.

narthur (commented):
Great work!

src/components/organisms/cardSlider/section/index.module.scss
```diff
@@ -19,3 +19,8 @@
 		text-decoration: none;
 	}
 }
+
+.hidden {
```
narthur, line 23: Do we still need this class?

narthur (commented):
lgtm! 🚀

src/components/organisms/cardSlider/section/index.tsx
```diff
@@ -90,18 +108,38 @@ export default function Section<T extends GraphqlInfiniteQuery, N>({
...
+		if (renderFail) {
+			failureCallback?.();
+		}
+	}, [renderFail, failureCallback]);
+
+	if (renderFail) {
 		return <></>;
 	}
 
+	const loadingCards: JSX.Element[] = [];
```
alangumer, line 135: Not used.

src/pages/[language]/discover/index.ts
```diff
@@ -1,34 +1,27 @@
...
-import serializableDehydrate from '~src/lib/serializableDehydrate';
+import { storeRequest } from '~src/lib/api/storeRequest';
+import getDehydratedProps, {
+	DehydratedProps,
+} from '~src/lib/getDehydratedProps';
 
 export default Discover;
 
-export async function getStaticProps({
+export async function getServerSideProps({
```
Comment (alangumer) [inline src/pages/[language]/discover/index.ts:14]: Why was this changed to server side?

alangumer (changes requested):

src/pages/[language]/discover/index.ts
Comment (jakecoble) [reply src/pages/[language]/discover/index.ts:14]: Featured Teachings was appearing and disappearing during page load if the user was logged in. It didn't look great, and some of the other solutions we came up with were awkward, so we moved the rendering server side.

Comment (alangumer) [reply src/pages/[language]/discover/index.ts:14]: I see, this is one of the main pages, so it gets a lot of loads, so I'd suggest to keep it static to take advantage of Vercel page caching.

src/components/organisms/cardSlider/section/index.tsx
```diff
@@ -90,15 +103,21 @@ export default function Section<T extends GraphqlInfiniteQuery, N>({
...
 				fetchNextPage();
 			}
 		},
-		[fetchNextPage]
+		[fetchNextPage, isFetchingNextPage, hasNextPage]
 	);
 
 	// Check if there's content to render, if not, return an empty JSX element
-	if (cards.length < 1) {
+	const renderFail = cards.length < 1;
```
alangumer, line 118: This is only used once, move it to the if.

src/containers/discover/index.tsx
```diff
@@ -9,14 +9,58 @@ import StorySeasons from '~components/organisms/cardSlider/section/storySeasons'
...
+	isLoading: boolean;
+	count: number | undefined;
+	isLoadingCount: boolean;
+	isServerSide: boolean;
+}): 'nothing' | 'featured' | 'continueListening' {
+	if (options.isServerSide) return 'nothing';
+	if (!options.isUserLoggedIn && !options.isLoading) return 'featured';
+	if (options.isLoading || options.isLoadingCount) return 'nothing';
+	if (options.isUserLoggedIn && !options.isLoadingCount) {
+		if (options.count ?? 0 > 0) return 'continueListening';
```
alangumer, line 34: There is no need to use ?? since undefined is not greater than 0.

alangumer (approved):

src/containers/discover/index.tsx
jakecoble: The TypeScript compiler complains if we don't ensure that `options.count` is a number.

`[checkTs] src/containers/discover/index.tsx(34,7): error TS18048: 'options.count' is possibly 'undefined'.`

PR #570, opened 2024-08-16
AV-1153: Fix prefetching and hydration

jakecoble:
Jira story: https://audioverse7.atlassian.net/jira/software/projects/AV/boards/3?assignee=712020:e12535da-4ed4-4700-9b40-8821b41cbf9e&selectedIssue=AV-1153

### DIFF
--- a/src/lib/hooks/useIsAuthenticated.ts
+++ b/src/lib/hooks/useIsAuthenticated.ts
@@ -1,22 +1,13 @@
-import { UseQueryResult } from '@tanstack/react-query';
+import { useGetIsAuthenticatedQuery } from './__generated__/useIsAuthenticated';
 
-import {
-	GetWithAuthGuardDataQuery,
-	useGetWithAuthGuardDataQuery,
-} from '~components/HOCs/__generated__/withAuthGuard';
-import { getSessionToken } from '~lib/cookies';
-
-export default function useIsAuthenticated(): UseQueryResult<
-	GetWithAuthGuardDataQuery,
-	unknown
-> & {
+export default function useIsAuthenticated(): {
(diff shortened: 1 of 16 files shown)


src/lib/hooks/useIsAuthenticated.ts
```diff
@@ -0,0 +1,13 @@
+import { useGetWithAuthGuardDataQuery } from '~components/HOCs/__generated__/withAuthGuard';
```
narthur, line 1: Since this graphql query is associated with the hook instead of the HOC now, let's move it to useIsAuthenticated.graphql and call the query something like `getIsAuthenticated`.

narthur (changes requested):
Great work!

jakecoble:
Steps used to identify missing prefetches:

1. Open site in a new window
2. Block https://graphql.audioverse.org/graphql
3. Scroll through each page, comparing two versions of the site: one with the graphql endpoint blocked and one with the endpoint unblocked.
4. Ensure those sections are fetched in prefetchQueries in that page's getServerSideProps and getStaticProps functions.

Steps used to find fetches that weren't using the generated queries:

1. Grep codebase for @tanstack/react-query imports.
2. Modify files using react-query directly to use the generated query code.

src/lib/hooks/useIsAuthenticated.ts
jakecoble: Should be done in the latest commit.

narthur:
Great work, @jakecoble!

narthur (approved):
Great work!

alangumer (approved):
Great work!

PR #591, opened 2024-10-24
Add favorites to continue listening carousel

jakecoble:


alangumer (changes requested):
The app doesn't have an infinite scroll, and takes into account duplicates from downloads and favorites, take a look at audioverse-mobile/src/containers/discover/index.tsx.

narthur:
This PR is no longer needed since AV agreed to no longer include favorites in the continue listening carousel.

https://audioverse7.atlassian.net/jira/software/projects/AV/boards/3?assignee=712020:e12535da-4ed4-4700-9b40-8821b41cbf9e&selectedIssue=AV-1356

PR #610, opened 2024-12-26
Bible UI (AV-1382, AV-1320)

narthur:
Stories:

- [AV-1382](https://audioverse7.atlassian.net/issues/AV-1382?filter=10009)
- [AV-1320](https://audioverse7.atlassian.net/issues/AV-1320?filter=10009)

[AV-1382]: https://audioverse7.atlassian.net/browse/AV-1382?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AV-1320]: https://audioverse7.atlassian.net/browse/AV-1320?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lcflight (approved):
Looking good.

scripts/save-graphql-version-indices.main.ts
```diff
@@ -0,0 +1,75 @@
+import fs from 'fs';
```
alangumer, line 1: This won't work for the Bible "King James Version (AudioVerse)" because new content is being added and a new deployment will be necessary to get the new data. Get the data dynamically as needed, for instance for the versions dropdown just the id and title are necessary, and for the books UI, just the sequences for that Bible.

alangumer (changes requested):

alangumer (changes requested):
This is the first pass. Also, there was an error when loading directly en/bibles/ENGKJV2/MAT/3, it looks that the issue is that before it was using the book id and now the book name.

src/services/bibles/getBible.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export default async function getBible(
+	versionId: string,
+): Promise<GraphqlBible | undefined> {
+	const fcbhMatch = getFcbhBible(versionId);
+
+	if (fcbhMatch) {
+		await Promise.all(
+			fcbhMatch.books.map(async (book) => {
+				book.chapters_full = await fetchFcbhChapters(
```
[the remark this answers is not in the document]
narthur: Hmm, I'm going to push back on this one.

`fetchFcbhChapters` uses `fetchResponse`, which in turn uses `manageAsyncFunction`. This means all calls to the FCBH API are memoized, so the referenced code shouldn't result in many duplicate calls to FCBH. We need the data from the fileset in order to determine id, title, and duration of each chapter in `transformVersion`. We have a separate hook, `useAudioUrlRefresh`, which handles refreshing stale FCBH audio URLs on the client.

The current design's goal is to restrict as much of our code as possible from needing to know about FCBH. I'm hopeful in the future we'll find a way for the GraphQL API to be able to proxy FCBH so the public site doesn't need to know anything about the two different audiobible sources. The hope is that handling Bibles the way I have will make that switch as easy as possible whenever that time comes.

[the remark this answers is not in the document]
alangumer: I see. The getBible function is used in bibles/[version].tsx and bibles/[version]/[book]/[chapter], the first route doesn't need the full details and URLs for all the books, just the books that are available, names and ids, so there is no need to call the API, the second route same as the first one, plus detailed information about the book and chapter URL, which is part of the fileset, so it should do just one API call to get the fileset for that book.

alangumer (changes requested):
Keep the book_id in the URL as it is in production; this allows us to standardize and use the same book_id across languages, also, the app relies on the book_id in the URL for deep linking.

jlaverde77 (commented):
I feel like the CI changes should be separate as this is quite a large change and we have the AV-1514 for it. I'm making the updates in a separate branch and PR and an additionalupdate to e2e.yml file.

src/services/bibles/fcbh/fetchFcbhBibles.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export const FCBH_VERSIONS = [
+	{
+		id: 'ENGKJV2',
+		title: 'King James Version (Dramatized)',
+		abbreviation: 'KJV',
+	},
+];
+
+export async function fetchFcbhBibles(): Promise<IBibleVersion[] | null> {
```
alangumer, line 13: Not used, remove it.

src/containers/bible/chapter.tsx
```diff
@@ -0,0 +1,290 @@
+import React, { useEffect, useRef, useState } from 'react';
```
alangumer, line 1: The GTM push event recording_view is misssing, take a look at recording.tsx.

src/services/bibles/fcbh/fetchFcbhChapters.ts
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	const fcbhId = bookId.split('/').pop() ?? '';
+	const filesets = response.data?.filter(({ book_id }) => book_id === fcbhId);
+
+	filesets.sort((a, b) => a.chapter_start - b.chapter_start);
+
+	return filesets.map(
+		({ book_id, book_name, chapter_start, path, duration }) => ({
+			id: `${book_id}/${chapter_start}`,
```
alangumer, line 31: The fileset id is missing in the id: `${bible.fileSetId}/${chapter.book_id}/${chapter.chapter_start}`.

src/services/bibles/transforms/chapterTransforms.ts
```diff
@@ -0,0 +1,121 @@
...
+		shareUrl: canonicalUrl,
+		recordingContentType: RecordingContentType.BibleChapter,
+		collection: {
+			id: chapter.version_id,
+			title: chapter.version_name,
+			contentType: CollectionContentType.BibleVersion,
+		},
+		speakers: [],
+		sponsor: { title: 'Faith Comes By Hearing' },
+		sequence: null,
```
alangumer, line 61: Specify at least the sequence title, which is the book name, this is used for playback tracking purposes on GTM, take a look at transformBibleBrainChapter on the app.

src/services/bibles/transforms/chapterTransforms.ts
```diff
@@ -0,0 +1,121 @@
...
+	if (!chapter.text) {
+		throw new Error('Text is required');
+	}
+
+	const canonicalPath = buildCanonicalPath(chapter);
+	const canonicalUrl = `https://www.audioverse.org${canonicalPath}`;
+
+	return {
+		id: chapter.id,
+		title: chapter.title,
```
alangumer, line 47: Only the first letter of each word is capitalized, take a look at transformBibleBrainChapter on the app.

alangumer (changes requested):

src/containers/bible/chapter.tsx
alangumer: The id should include the FCBH fileset id, and not the custom Bible id.

src/services/bibles/fcbh/fetchFcbhChapters.ts
alangumer: The id should include the FCBH fileset id, and not the custom Bible id.

src/services/bibles/transforms/chapterTransforms.ts
alangumer: When playing a chapter, the video element data attribute data-series in andMiniplayer.tsx doesn't have a value.

src/containers/bible/chapter/useRefreshedRecordings.graphql
```diff
@@ -0,0 +1,11 @@
+fragment Refreshable on Recording {
```
alangumer, line 1: Remove this because you are not using it as a GraphQL fragment but as a TS type.

alangumer (changes requested):

alangumer (approved):

PR #612, opened 2025-01-06
AV-1434-Change-Home-page-content-with-Abe-s-update-Website

jlaverde77:


src/components/molecules/button.module.scss
```diff
@@ -48,6 +48,22 @@ $button-height: 31px;
 	}
 }
 
+.superHero {
```
alangumer, line 51: There is no need to create a new button type, the only difference is the font size, use super and just create a class with the font size style. You might not even need to create a new class since the font-size looks bigger than the mockup, and so you could just use super.

alangumer (changes requested):
Remove the old images, and convert the new image to JPG since is smaller for web, and translate the content to the other languages.

src/components/molecules/button.module.scss
jlaverde77: Unfortunately, Abe and John specifically asked for it to be bigger. I'll change it back to super and add the font size change.

jlaverde77:
Based on the Team Huddle decision on this, this will be closed without merging.

PR #631, opened 2025-04-17
Update E2E workflow for firewall compat

narthur:


src/components/atoms/lineHeading.module.scss
```diff
@@ -2,7 +2,7 @@
 
 .heading {
 	display: flex;
-	color: $ts-red;
+	color: $ts-red-on-white;
```
jlaverde77, line 5: This component is used in production. Why is it being changed here?

src/components/molecules/button.module.scss
```diff
@@ -39,8 +39,8 @@ $button-height: 31px;
 	&:hover {
 		color: $ts-white;
 	}
-	background: $ts-red;
-	border: 1px solid $ts-red;
+	background: $ts-red-under-white;
```
jlaverde77, line 42: This as well

jlaverde77 (commented):
Are the color changes something leftover from Bible update?

src/components/atoms/lineHeading.module.scss
narthur: Reverted

src/components/molecules/button.module.scss
narthur: Reverted

narthur:
> Are the color changes something leftover from Bible update?

The accessibility check was failing. Before realizing it was due to a bug with pa11y-ratchet not respecting ignored codes, I first tried fixing the reported issues.

I've now reverted the out-of-scope accessibility fixes.

PR #635, opened 2025-05-08
improve(Copyright): adjust fine print for EGW audiobooks

Keshavdulal:


Keshavdulal:
@jlaverde77 the test seems to be failing on `master` too.
It was passing when I committed and wasn't flagged by pre-commit check.
image

